PR #2, opened 2026-06-11
Fix single-env service restart reloading unrelated gateways

### LINKED ISSUE #1: Bug: single-env service restart reloads unrelated managed gateways
## Summary

`ocm service restart <env>` is documented and expected to restart only the targeted env gateway, but on a multi-env OCM host a targeted restart caused the shared OCM supervisor to SIGTERM multiple unrelated managed gateways.

This interrupted active agent turns in other envs.

## Observed Evidence

Local runtime: `ocm 0.2.21` / repo `v0.2.21` (`6547714`).

An agent in one managed env installed/configured an OpenClaw plugin, then ran a targeted service restart for that env. Shortly afterward, unrelated managed gateways also received SIGTERM and active Codex turns failed with `codex app-server client closed before turn completed`.

OCM daemon stderr showed a fan-out reload rather than only the targeted env:

```text
ocm service: reloading <unrelated-env-a> (runtime:local-dev)
ocm service: reloading <unrelated-env-b> (runtime:stable)
ocm service: reloading <target-env> (runtime:stable)
ocm service: starting <unrelated-env-a> (runtime:local-dev)
ocm service: starting <unrelated-env-c> (launcher:...)
ocm service: starting <unrelated-env-b> (runtime:stable)
ocm service: starting <target-env> (runtime:stable)
```

An unrelated gateway log confirmed shutdown:

```text
[gateway] signal SIGTERM received
[gateway] received SIGTERM; shutting down
```

The interrupted agent log confirmed the user-visible failure:

```text
[agent/embedded] codex app-server client closed before turn completed
```

## Expected Behavior

- `ocm service restart <env>` should restart only the targeted env.
- `ocm @<env> -- doctor --fix` should mutate/check only the targeted env.
- Other managed envs should remain running unless their own service policy or launch spec changes.

## Diagnosis

`SupervisorChildSpec` includes `process_env`, which was derived from the caller environment. The daemon reconciler restarts any running child whose stored spec differs from the next generated spec.

That means when a different agent/session/LaunchAgent rewrites the shared supervisor state, unrelated children can appear changed because caller-only environment variables changed, even if their actual runtime binding did not.

Likely problematic caller-only fields include values like `PWD`, `XPC_SERVICE_NAME`, `GH_TOKEN`, Codex/session-specific vars, and volatile `PATH` entries.

There is a second restart-specific trigger: the old `ocm service restart <env>` implementation briefly wrote a global state where the target env was stopped, then immediately wrote a global state where it was running again. That transient global desired-state rewrite gave unrelated children another chance to look changed and be reloaded.

## Impact

A normal targeted service operation in one env can interrupt live conversations and delivery in other envs, defeating the main OCM isolation guarantee.

## Proposed Fix

PR #2 fixes this with two changes:

- Build supervised gateway child environments from a stable allowlist instead of inheriting arbitrary caller env vars. Keep env-specific `OPENCLAW_*` / `OCM_*` values plus stable service/runtime keys and proxy/cert vars, and sanitize volatile `PATH` entries.
- Make `ocm service restart <env>` use a targeted supervisor restart request for the selected child instead of temporarily removing the env from global desired state.

hannesrudolph:
## Summary

Fixes #1.

This fixes two supervisor behaviors that could make a targeted `ocm service restart <env>` restart unrelated managed gateways:

- supervised gateway child specs now keep only stable environment keys instead of inheriting arbitrary caller/session context
- `ocm service restart <env>` now writes a targeted restart request for the selected child instead of temporarily removing that env from the shared desired-state file

## Root Cause

The OCM daemon reconciles running children by comparing the stored `SupervisorChildSpec` with the next generated spec. Those specs included `process_env`, which was previously derived from the caller environment. When a different agent session or LaunchAgent rewrote supervisor state, unrelated child specs could differ because of caller-only values such as `PWD`, `XPC_SERVICE_NAME`, `GH_TOKEN`, Codex/session variables, or volatile `PATH` entries.

The restart command also used environment metadata as a control channel: it briefly marked the target env stopped, synced global supervisor state, then marked it running again. During that global state transition, the daemon could reload siblings whose regenerated specs looked different.

## What Changed

Supervised gateway child environments are now built from a stable allowlist:

- env-specific `OPENCLAW_*` / `OCM_*` values
- stable service/runtime keys: `HOME`, `PATH`, `TMPDIR`, `SHELL`, `OCM_HOME`, `OCM_SELF`
- proxy and certificate-related service keys already recognized by the supervisor

The retained child `PATH` is also sanitized to drop volatile Codex temp paths and direct Homebrew Cellar Node paths before the spec is written.

For already-running services, `ocm service restart <env>` now appends a one-child restart request to supervisor state. The daemon consumes and clears that request, stops only the targeted running child, and queues only that child to restart. The old start/sync path is still used when the env is not already enabled and running.

## Validation

- `cargo test`
- `cargo build --release`
- Installed locally and restarted the OCM daemon
- Live verification: `ocm service restart rescue` changed only `rescue` from PID `71696` to `72723`; `clawpatrol`, `main`, and `openclaw` stayed at `71692`, `71693`, and `71694`

### DIFF
--- a/src/service/manage.rs
+++ b/src/service/manage.rs
@@ -92,21 +92,67 @@ pub fn restart_service(
     env: &BTreeMap<String, String>,
     cwd: &Path,
 ) -> Result<ServiceActionSummary, String> {
+    ensure_gateway_binding(name, env, cwd)?;
     let env_service = EnvironmentService::new(env, cwd);
     let meta = env_service.get(name)?;
-    if meta.service_enabled && meta.service_running {
-        env_service.set_service_policy(name, Some(true), Some(false))?;
+    if !(meta.service_enabled && meta.service_running) {
+        return update_service(
+            name,
+            "restart",
+            Some(true),
+            Some(true),
+            false,
+            ServiceSupervisorPolicy::EnsureRunning,
+            env,
+            cwd,
+        );
     }
-    update_service(
-        name,
-        "restart",
-        Some(true),
-        Some(true),
-        true,
-        ServiceSupervisorPolicy::EnsureRunning,
-        env,
-        cwd,
-    )
+
+    let before = super::inspect::service_status_fast(name, env, cwd)?;
+    let Some(previous_pid) = before.child_pid else {
+        return update_service(
+            name,
+            "restart",
+            Some(true),
+            Some(true),
+            false,
+            ServiceSupervisorPolicy::EnsureRunning,
+            env,
+            cwd,
+        );
+    };
+    if !(before.ocm_service_running && before.running) {
+        return update_service(
+            name,
+            "restart",
+            Some(true),
+            Some(true),
+            false,
+            ServiceSupervisorPolicy::EnsureRunning,
+            env,
+            cwd,
+        );
+    }
+
+    let supervisor = SupervisorService::new(env, cwd);
+    let request_id = supervisor.request_child_restart(name)?;
+    let restart_result = wait_for_restart_action_summary(name, previous_pid, env, cwd);
+    let clear_result = supervisor.clear_child_restart_request(name, &request_id);
+    match (restart_result, clear_result) {
+        (Ok((summary, warnings)), Ok(())) => {
+            Ok(service_action_summary("restart", summary, warnings))
+        }
+        (Ok((summary, mut warnings)), Err(clear_error)) => {
+            warnings.push(format!(
+                "restart completed, but failed to clear restart request: {clear_error}"
+            ));
+            Ok(service_action_summary("restart", summary, warnings))
+        }
+        (Err(restart_error), Ok(())) => Err(restart_error),
+        (Err(restart_error), Err(clear_error)) => Err(format!(
+            "{restart_error}; additionally failed to clear restart request: {clear_error}"
+        )),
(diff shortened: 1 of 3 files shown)


shakkernerd:
Merged using squash.

What changed after review:
- Targeted service restart now refreshes and restarts only the requested env child.
- Stopped desired-running children are requeued instead of returning a no-op restart.
- Stable runtime env and PATH entries are preserved while volatile caller state is stripped.

Verification:
- Refreshed against current `origin/main` before pushing and merging.
- Local CI commands passed: `cargo fmt --check` and `cargo test`.
- Live isolated E2E passed: target child restarted with a new PID while sibling child kept its PID.
- Auto-review: clean after accepted findings were resolved.
- GitHub checks: no required checks configured / no checks reported on the PR branch.

Landed on `main`:
- Final landed SHA: 07cf0d48fb530433a91c7995bd792faf7c9c77f2; range 6547714a371b59ea8b4f9efddd3d7382d83e8c3f...07cf0d48fb530433a91c7995bd792faf7c9c77f2.

Thanks @hannesrudolph!

PR #3, opened 2026-06-20
Fix restart recovery when supervisor misses a child restart

hannesrudolph:
## Summary

When `ocm service restart <env>` cannot observe a replacement child within its wait window, requeue only that child's durable restart request and recheck the OCM supervisor.

## Failure mode

A supervisor or gateway interruption can occur after the restart request is written but before a new child PID is observed. Recovery must retain the targeted-restart isolation from #2: rebuilding global supervisor state can reload unrelated gateways.

## What changed

- Reissue only the target child's restart request when restart observation times out.
- Activate the OCM supervisor without running a global state sync.
- Preserve persisted sibling child specs during recovery.
- Report clearly whether targeted supervisor recovery succeeded.
- Add a two-child regression test proving recovery cannot rewrite the sibling spec.

## Validation

- `cargo fmt --all --check`
- `cargo test --test daemon_runtime_tests --test service_command_tests -- --test-threads=1` (41 passed)
- Both unrelated timing failures from an earlier parallel run passed individually.
- Autoreview: clean, no actionable findings, confidence 0.88.

### DIFF
--- a/src/service/manage.rs
+++ b/src/service/manage.rs
@@ -141,12 +141,12 @@ pub fn restart_service(
                     ));
                 }
             } else {
-                match supervisor.ensure_daemon_running() {
+                match supervisor.recover_child_restart(name) {
                     Ok(_) => status.warnings.push(
-                        "gateway restart was not observed; restart request was retained and the supervisor was rechecked".to_string(),
+                        "gateway restart was not observed; the targeted restart request was requeued and the supervisor was rechecked".to_string(),
                     ),
                     Err(error) => status.warnings.push(format!(
-                        "gateway restart was not observed; restart request was retained, but supervisor recovery failed: {error}"
+                        "gateway restart was not observed; targeted supervisor recovery failed: {error}"
                     )),
                 }
             }
--- a/src/service/manage.rs
+++ b/src/service/manage.rs
@@ -140,6 +140,15 @@ pub fn restart_service(
                         "restart completed, but failed to clear restart request: {clear_error}"
                     ));
                 }
+            } else {
+                match supervisor.ensure_daemon_running() {
+                    Ok(_) => status.warnings.push(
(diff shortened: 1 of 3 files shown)


vincentkoc (changes requested):
Blocking: the timeout recovery path reintroduces global sibling reconciliation.

`restart_service()` now calls `ensure_daemon_running()` when the target restart is not observed (`src/service/manage.rs:143`). That method begins with `self.sync()` (`src/supervisor/mod.rs:326`), which rebuilds and writes every managed child spec. This defeats the targeted-restart protection from #2: `request_child_restart()` deliberately preserves persisted sibling specs, but the new timeout branch immediately replaces them from the current caller environment. The daemon then treats any sibling spec difference as `needs_restart` and reloads that unrelated child (`src/supervisor/mod.rs:1009`).

I reproduced this on the PR head with two children. After `request_child_restart("rescue")`, the persisted `main` spec correctly retained its old environment. Calling `ensure_daemon_running()` with a changed stable runtime variable updated `main` too, which makes it eligible for an unrelated restart. The throwaway proof test passed, confirming the side effect; it was removed afterward. The submitted suites also pass: `cargo test --test daemon_runtime_tests --test service_command_tests` (40/40), so existing coverage does not catch this path.

Please separate daemon liveness/recovery from global state sync (for example, a no-sync ensure/recover operation), or refresh/reissue only the target restart request while preserving sibling specs. Add a two-child timeout/recovery test asserting the target is retried and the sibling spec/PID remains unchanged.

vincentkoc (approved):
The blocking global-sync regression is fixed at 326e329. Targeted recovery now requeues only the requested child, preserves sibling specs, and activates the supervisor without a global sync. Local serial proof, exact-head CI, and autoreview are clean.

PR #15, opened 2026-07-12
fix(logs): preserve follow integrity across rotation

vincentkoc:
## What Problem This Solves

`ocm logs --follow` could corrupt or lose record boundaries around rotation, truncation, missing-path gaps, split UTF-8 writes, and delayed writes to renamed files. The previous implementation also rescanned the full file prefix during steady-state polling, making follow cost grow with log size.

## Why This Change Was Made

The follow path now uses a byte-oriented cursor with stable file identity and retained file handles across bounded rotation windows. It reads only the requested startup tail or newly appended range, keeps multiline and partial records intact, and explicitly separates records when the underlying source changes.

Rotation handling retains at most four old handles for at most four quiet polls. A same-identity source returning after a missing-path gap is reattached before draining, while replacement sources and independent retired writes remain separated for both single-stream and merged output. Windows file identity uses volume serial and file index information.

## User Impact

- Followed logs no longer lose or concatenate records during rename/create rotation, truncation, or delayed writer handoff.
- Split UTF-8 and multiline records remain byte-exact until complete.
- Merged stdout/stderr startup ordering and tail selection are global and deterministic.
- Steady-state follow work is bounded: idle polls read zero bytes, growth reads only the appended range, and a 2.5 MB tail fixture reads at most 8 KiB at startup.

## Evidence

- `cargo test --lib logs --no-fail-fast`: 43 passed
- `cargo test --test logs_command_tests --no-fail-fast`: 7 passed
- `cargo test --no-fail-fast`: full suite passed
- strict `cargo clippy --lib --tests -- -D warnings` with the repository's existing lint allowances: passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- focused Sol/high autoreview thread `019f54f1-d6b2-7d40-aa89-6e273183b18c`: no actionable findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,3 +24,4 @@ All notable changes to OCM are documented here.
 - Make release preparation rollback-safe, verify signed annotated tags on every resume path, and publish branch and tag refs atomically.
 - Build locked release artifacts only from a GitHub-verified tag matching the package version, require the complete supported target matrix before publication, publish a versioned checksum-verifying installer, pin CI actions, and test the declared Rust 1.88 minimum.
 - Preserve uncommitted dev worktrees and reject unrelated checkout collisions by requiring exact Git worktree registration before reuse or removal.
+- Make single and merged log following rotation-safe and byte-safe, preserving split or invalid UTF-8, multiline record ordering, and boundaries around unterminated records.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -396,6 +396,7 @@ dependencies = [
  "time",
  "unicode-width",
  "ureq",
+ "windows-sys 0.61.2",
  "zip",
 ]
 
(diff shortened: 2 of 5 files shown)


vincentkoc:
Land-ready verification for exact head `c28e09a8bc46e8c84f2270e8ce046ebe363a67a2`:

- `cargo test --lib logs --no-fail-fast`: 43 passed
- `cargo test --test logs_command_tests --no-fail-fast`: 7 passed
- `cargo test --no-fail-fast`: full suite passed
- strict `cargo clippy --lib --tests -- -D warnings` with repository lint allowances: passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- focused Sol/high autoreview: no actionable findings
- exact-head GitHub Actions CI run 29182316232: Format, Rust 1.88 minimum, Ubuntu tests, and macOS tests all passed

Merged without squash as `8999f10dc91010b52180338643fd601192f76e07`. Exact-main CI run 29182409092 also passed all four jobs, including Ubuntu and macOS install smoke tests.

The implementation remains bounded to four retained handles and four quiet polls, reads zero bytes on idle polls, and reads only appended ranges during normal growth.

PR #16, opened 2026-07-12
fix(dev): serialize and own source watch lifecycle

vincentkoc:
## What Problem This Solves

`ocm dev --watch` could overlap another watch for the same environment, publish stale ownership metadata, or leave the background service and source process tree in the wrong state after failures, signals, job-control suspension, or wrapper exits.

That made repeated development and retesting unsafe: a second watch could race the first, detached descendants could survive cleanup, and service takeover could be restored by the wrong owner.

## Why This Change Was Made

- Serialize source-watch ownership with a per-environment lock and lease generation.
- Publish and validate process identity before treating metadata as authoritative.
- Own and terminate the full source-watch process tree on Unix and Windows.
- Assign Windows watchers to a Job Object before publishing the lease.
- Preserve interactive stdin, terminal foreground ownership, and `Ctrl-Z` / `fg` job control.
- Keep service restoration under the active lease and restore the previous policy on every failure or termination path.
- Remove the unleased source-watch override mutation API.
- Reclaim stale metadata without allowing concurrent readers to revive it.

## User Impact

Only one source watch can own an environment at a time. Stale watches are reclaimed, active watches cannot be displaced by stale metadata, terminal suspension/resume behaves like a normal foreground job, and forced service takeover is restored reliably when the watch exits or cannot start.

## Evidence

Reviewed head: `7b114254056022627f919ef27c61cd03cc524afa`

- `cargo fmt --check`
- `cargo check`
- `cargo test --test dev_command_tests --test source_watch_tests` (49 passed)
- `cargo test` (full suite passed)
- Windows API compile probe for the Job Object/process APIs passed; full Windows cross-check was unavailable locally because `x86_64-w64-mingw32-gcc` is not installed.
- Fresh whole-branch Sol/high autoreview: no findings; patch correct with 0.84 confidence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,4 +25,4 @@ All notable changes to OCM are documented here.
 - Build locked release artifacts only from a GitHub-verified tag matching the package version, require the complete supported target matrix before publication, publish a versioned checksum-verifying installer, pin CI actions, and test the declared Rust 1.88 minimum.
 - Preserve uncommitted dev worktrees and reject unrelated checkout collisions by requiring exact Git worktree registration before reuse or removal.
 - Make single and merged log following rotation-safe and byte-safe, preserving split or invalid UTF-8, multiline record ordering, and boundaries around unterminated records.
-- Prevent overlapping source watches, remove the unleased override mutation path, reclaim stale watch locks, and restore background-service policy when takeover cannot stop or exits with an error.
+- Prevent overlapping source watches, remove the unleased override mutation path, reclaim stale watch locks, handle termination signals, and restore background-service policy when takeover cannot stop or exits with an error.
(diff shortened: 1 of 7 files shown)


vincentkoc:
Land-ready verification for exact head `7b114254056022627f919ef27c61cd03cc524afa`:

- `cargo fmt --check`
- `cargo check`
- `cargo test --test dev_command_tests --test source_watch_tests` (49 passed)
- `cargo test` (full local suite passed)
- Fresh whole-branch Sol/high autoreview: no findings
- GitHub CI run https://github.com/shakkernerd/ocm/actions/runs/29182653933
  - Format: passed
  - Rust 1.88 minimum: passed
  - Ubuntu test suite and install smoke: passed
  - macOS test suite and install smoke: passed

Known platform gap: the Windows API surface was compile-probed locally, but a full local Windows cross-check was unavailable because the MinGW C compiler is not installed.

PR #18, opened 2026-07-12
fix(releases): preserve overlapping channel selectors

vincentkoc:
## What Problem This Solves

When npm points multiple recognized dist-tags at the same OpenClaw version, OCM currently keeps only the highest-priority channel. For example, if both `latest` and `beta` reference one version, selecting `beta` fails even though npm advertises it.

## Why This Change Was Made

OCM now retains every normalized channel associated with a published version while preserving the existing stable-first canonical channel for unfiltered and version-scoped views. Channel-scoped selection matches the retained aliases and returns the channel the operator requested, keeping runtime provenance and follow-up commands accurate.

This intentionally excludes the unrelated metadata validation, version ordering, dependency, and CLI changes from the original aggregate audit commit.

## User Impact

`release`, `runtime`, install, and upgrade paths can resolve `stable`, `beta`, or `dev` correctly even when two dist-tags temporarily point to the same package version during promotion. Existing JSON output remains unchanged because the internal alias list is not serialized.

## Evidence

- `cargo test --locked --test release_manifest_tests` (9 passed)
- `cargo test --locked --test release_command_tests` (8 passed)
- `cargo test --locked` (full suite passed)
- `git diff --check`
- Sol/high autoreview: clean, no accepted/actionable findings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,3 +30,4 @@ All notable changes to OCM are documented here.
 - Preserve working runtimes until replacement artifacts, dependencies, and metadata are ready; require release integrity and reject unsafe artifact names.
 - Write private service definitions, preserve existing service-directory modes, escape systemd specifiers, reject directive injection, and fail closed on unknown launchd users or cross-store service ownership.
 - Serialize supervisor restart requests, restore service policy after failed reconciliation, preserve raw warnings, and fail upgrades when a gateway restart cannot be observed.
+- Preserve every recognized npm release channel when multiple dist-tags point to the same OpenClaw version.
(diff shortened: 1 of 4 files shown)


vincentkoc:
Land-ready verification for exact head `3fb2372a010bd39a2e972a2af494a4158b33b0a3`:

- `cargo test --locked --test release_manifest_tests` - 9 passed
- `cargo test --locked --test release_command_tests` - 8 passed
- `cargo test --locked` - full suite passed
- `cargo fmt --all -- --check` - passed
- `git diff --check` - passed
- Sol/high autoreview - clean, no accepted/actionable findings
- GitHub Actions run `29183375788` - Format, Rust 1.88 minimum, Ubuntu tests, and macOS tests all passed

Behavior proof: a package version advertised by both `latest` and `beta` retains both selectors; `latest` resolves it as `stable`, while `beta` resolves the same version with beta provenance. No known proof gaps.

PR #21, opened 2026-07-12
test(release): derive tag fixture from package version

vincentkoc:
## What Problem This Solves

The release-tag verification test hardcoded `v0.2.25`, so the `v0.2.26` version bump made both Linux and macOS `main` CI fail even though the release code was unchanged.

## Why This Change Was Made

Derive the valid fixture tag from `CARGO_PKG_VERSION`. Keep the mismatch case pinned to an obviously different version so the negative contract remains explicit.

## User Impact

Future version bumps no longer require a manual test-fixture update or leave release `main` red.

## Evidence

- `cargo fmt --all --check`
- `cargo test --locked --test release_asset_tests` (6/6)
- Sol/high autoreview: clean, confidence 0.97
- Reproduced current `main` failure in Actions run 29203432582

### DIFF
--- a/tests/release_asset_tests.rs
+++ b/tests/release_asset_tests.rs
@@ -207,6 +207,7 @@ esac
 fn verify_release_tag_requires_a_verified_annotated_tag_matching_the_package() {
     let root = TestDir::new("verify-release-tag");
     let fake_bin = root.child("bin");
+    let package_tag = format!("v{}", env!("CARGO_PKG_VERSION"));
     fs::create_dir_all(&fake_bin).unwrap();
     let expected_commit_output = Command::new("git")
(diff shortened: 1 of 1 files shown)


vincentkoc:
Land-ready verification:

- Exact head: `d0a1b3630ff8fa614904d0f02f9ba51ac63d4c04`
- `cargo fmt --all --check`
- `cargo test --locked --test release_asset_tests` (6/6)
- Sol/high autoreview: clean, confidence 0.97
- GitHub CI: Format, Rust 1.88, Windows compile, Ubuntu tests, and macOS tests all green

This fixes the version-bump-only failure from main run 29203432582. Merging with a merge commit; no squash.

PR #23, opened 2026-07-12
fix(service): serialize daemon ownership lifecycle

vincentkoc:
## What Problem This Solves

OCM stores share one OS daemon identity. Concurrent lifecycle commands could race ownership checks, and stopping the last environment could leave the shared definition installed. Failure rollback could also strand ownership after a partially completed first start.

## Why This Change Was Made

- serialize daemon ownership, policy mutation, synchronization, and teardown under one lifecycle lock
- validate the complete expected service definition before teardown
- remove the shared daemon after the last desired-running environment stops or is uninstalled
- restore the prior installed, registered, and running daemon state on failures
- make launchd teardown confirm each unload attempt and keep systemd stop/uninstall idempotent

## User Impact

Multiple OCM stores can no longer steal or strand the shared daemon binding. Repeated stop/uninstall remains safe, foreign-owned definitions are preserved, and failed first starts release ownership for another store.

## Evidence

- `cargo fmt --all`
- `cargo test --test service_command_tests` (35 passed)
- `cargo test service::platform --lib` (18 passed)
- high-thinking `gpt-5.6-sol` autoreview: clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to OCM are documented here.
 
 ### Fixed
 
+- Reject an already-running shared OCM daemon when its service definition belongs to another `OCM_HOME`, and remove the shared service binding after the last running environment stops or is uninstalled.
 - Reuse an identical healthy official runtime when concurrent installers publish it first instead of failing the waiting command.
 - Match environment-owned processes by path components so destroy never targets sibling-prefix directories.
 - Replace copied OpenClaw config symlinks with environment-owned files without rewriting external targets.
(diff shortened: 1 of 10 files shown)


vincentkoc:
Landed as an exact fast-forward to main at https://github.com/shakkernerd/ocm/commit/f40132fa2b2b93085920d33a7bfb37538d799b43. GitHub disables both rebase and merge commits for this repository, while this release lane explicitly forbids squash, so direct fast-forward was the only history-preserving path. All five PR commits and their authorship are preserved. Exact-head CI passed on macOS, Ubuntu, Windows compile, formatting, and Rust 1.88: https://github.com/shakkernerd/ocm/actions/runs/29208333569

PR #34, opened 2026-07-17
fix: preserve configured OpenClaw workspaces

hannesrudolph:
## Summary

- preserve OpenClaw workspaces derived from effective agent configuration across snapshots, exports, clones, imports, migrations, cleanup, and upgrade rollback
- resolve JSON5 includes, generated and explicit agent workspaces, and OpenClaw environment substitution without preserving unconfigured workspace-like runtime directories
- reject relative, external, symlink-escaping, non-regular, or include-owned workspace configurations before unsafe mutation

## Incident

An OCM-managed OpenClaw upgrade rollback restored only `.openclaw/workspace`. Secondary agent workspaces were absent from the snapshot and could be deleted during rollback.

The original prefix-based fix covered common `.openclaw/workspace-*` directories, but it did not cover arbitrary configured workspace paths and could preserve unrelated lookalike directories. This revision derives the durable workspace set from OpenClaw's effective configuration instead.

## Validation

- `cargo test --locked`
- `cargo clippy --workspace --all-targets --locked -- -D warnings -A clippy::question_mark -A clippy::large_enum_variant`
- `cargo fmt --check`
- targeted clone, import, migration, snapshot, cleanup, upgrade, include, environment substitution, path containment, and non-regular file regressions

### DIFF
--- a/src/env/health.rs
+++ b/src/env/health.rs
@@ -71,7 +71,7 @@ impl<'a> EnvironmentService<'a> {
     pub fn cleanup_preview(&self, name: &str) -> Result<EnvCleanupSummary, String> {
         let env = self.get(name)?;
         let config_audit = audit_openclaw_config(&env, &self.list()?);
-        let state_audit = audit_openclaw_state(&env, &self.list()?);
+        let state_audit = audit_openclaw_state(&env, &self.list()?, self.env);
         let doctor = self.doctor(name)?;
         let actions = cleanup_actions(&env, &doctor, &config_audit, &state_audit);
         Ok(build_cleanup_summary(&env, doctor, false, actions, None))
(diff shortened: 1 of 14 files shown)


shakkernerd:
Merged using rebase.

What changed:
- preserve OpenClaw workspaces derived from effective agent configuration across snapshots, exports, clones, imports, migrations, cleanup, and upgrade rollback
- reject unsafe external, relative, symlinked, non-regular, or include-owned workspace configurations before mutation
- preserve config includes and avoid treating workspace-like runtime directories as durable state

Verification:
- rebased onto current `main`
- full local test suite and all-target check passed
- required Format, Rust 1.88, Windows, Ubuntu, and macOS checks passed

Landed: `39699d3..0089e4d`

Thanks @hannesrudolph!

PR #37, opened 2026-07-20
fix: retry archives after transient file races

fuller-stack-dev:
## Summary

- retry an environment archive write once when an entry disappears after traversal has enumerated it
- restart the whole archive transaction instead of retrying one open or silently skipping the missing path
- keep root/output failures, permission errors, malformed links, and every non-`NotFound` I/O error fatal

## Incident

An OCM pre-upgrade snapshot failed while an active SQLite `-shm` file disappeared between directory enumeration and tar insertion. Stopping the environment avoided the race, but upgrades currently create their rollback snapshot while the environment may still be live.

## Design

The archive layer now classifies `NotFound` only for child-entry operations. The first such error discards the partial tar and reruns archive creation from a fresh traversal. A second disappearance fails normally.

Retrying the transaction is intentionally narrower and safer than reopening one path or skipping it. In particular, if a SQLite WAL disappears during a checkpoint, the retry recopies the database from the post-checkpoint tree rather than retaining an earlier database copy without its WAL. This does not attempt to turn live filesystem traversal into a SQLite backup protocol.

Stable traversal remains sorted and takes one pass. Only the observed race adds one bounded O(n) retry.

## Validation

- regression failed on the untouched base with `No such file or directory` after metadata and passes with the change
- focused archive tests: 4 passed
- source-blind CLI snapshot/restore proof: disappearing enumerated file passed; rebuilt-pass marker restored; unreadable stable file failed
- `cargo +1.88.0 check --workspace --all-targets --locked`
- `cargo fmt --check`
- `cargo clippy --workspace --all-targets --locked -- -D warnings -A clippy::question_mark -A clippy::large_enum_variant`
- `cargo test --locked -- --skip bin_wrapper_runs_with_an_overridden_home`
- full dev integration target rerun with a CI-like system Git PATH: 39 passed
- autoreview: clean, no accepted/actionable findings

The two allowed clippy lints are pre-existing warnings in untouched files. The locally skipped wrapper test requires rustup under the real home; GitHub CI provides that standard layout.

### DIFF
--- a/src/infra/archive.rs
+++ b/src/infra/archive.rs
@@ -13,6 +13,28 @@ use crate::env::{default_service_enabled, default_service_running};
 
 pub const ENV_ARCHIVE_METADATA_PATH: &str = "meta/env.json";
 pub const ENV_ARCHIVE_ROOT_DIR: &str = "root";
+const MAX_ENV_ARCHIVE_WRITE_ATTEMPTS: usize = 2;
+
+enum EnvArchiveWriteError {
+    EntryDisappeared(String),
+    Fatal(String),
+}
+
+impl EnvArchiveWriteError {
+    fn from_entry_io(error: io::Error) -> Self {
+        if error.kind() == io::ErrorKind::NotFound {
+            Self::EntryDisappeared(error.to_string())
+        } else {
+            Self::Fatal(error.to_string())
+        }
+    }
+
+    fn into_message(self) -> String {
+        match self {
+            Self::EntryDisappeared(message) | Self::Fatal(message) => message,
+        }
+    }
+}
 
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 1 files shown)


shakkernerd:
Merged using squash.

What changed after review:
- Rebased the contributor commit onto current `origin/main` with a content-neutral range diff.
- Preserved the bounded full-archive retry and fatal handling for persistent or non-`NotFound` failures.
- Preserved contributor authorship and added reviewer attribution.

Verification:
- Rebased/refreshed onto current `origin/main` before pushing.
- GitHub CI run 29780581713: format, Rust 1.88, Windows compile, Ubuntu tests, and macOS tests passed.
- Exact-head archive regression tests: 4 passed.
- Auto-review: exact-head fast review clean with no blocking findings.
- GitHub checks: all 5 required checks green.

Landed on `main`:
- Final landed SHA: 39699d3292228623e483fdfed8f43d9103125a60; range 00f2e65f7ec977420426b6db4d5a112780235666...39699d3292228623e483fdfed8f43d9103125a60.

Thanks @fuller-stack-dev!

PR #47, opened 2026-07-23
fix: isolate shared runtime upgrades

shakkernerd:
## Summary

- reject environment upgrades that would replace runtime bytes used by sibling environments
- serialize the final sibling check with runtime preparation to close the registry race
- preserve named runtime reuse and document exact-version isolation guidance

## Verification

- real OpenClaw `2026.6.11`, `2026.6.33`, and `2026.7.1` shared-runtime command flows
- complete 25-test upgrade suite
- full `cargo test`
- `cargo check --all-targets`
- Windows GNU cross-check
- `cargo fmt --check`
- `git diff --check`
- real `ocm help upgrade` output

### DIFF
--- a/src/cli/render/help.rs
+++ b/src/cli/render/help.rs
@@ -581,6 +581,7 @@ pub fn upgrade_help(cmd: &str) -> String {
             "Local-repo simulations validate deps/build/UI build before post-update checks.",
             "Post-update checks run update-mode doctor, plugin update dry-run, and gateway status.",
             "Channel-tracked runtimes move forward automatically.",
+            "An env upgrade will not replace runtime bytes shared with another env; use --runtime to reuse those bytes or an exact --version for an isolated target.",
             "Upgrades create a pre-upgrade snapshot before changing env state.",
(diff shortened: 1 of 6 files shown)


shakkernerd:
Merged using rebase.

What changed after review:
- Rejected environment upgrades that would replace runtime bytes used by sibling environments.
- Protected explicit shared targets while preserving named runtime reuse when no bytes are mutated.
- Serialized the final sibling check with runtime preparation and documented the safe alternatives.

Verification:
- Refreshed `origin/main` and verified the landed four-commit sequence and attribution.
- Ran the full test suite, all-target checks, and Windows compilation on AWS Crabbox lease `cbx_a55a7013c2d7`.
- Exercised real OpenClaw 2026.6.11, 2026.6.33, and 2026.7.1 runtime-sharing cases against the frozen source point.
- Auto-review was unavailable; focused exact-head manual review found no actionable issues.
- All five protected checks passed.

Landed: `d9fe13a100821f97b0a1c220895e0ef29bfd60df`
Range: `5004d13d881c3b9ddcec1872c8fac272034e70f5...d9fe13a100821f97b0a1c220895e0ef29bfd60df`

Thanks @shakkernerd!

PR #55, opened 2026-07-23
fix: serialize complete upgrade operations

shakkernerd:
## Summary
- hold the per-environment operation lock for the complete normal upgrade lifecycle
- use locked binding, service, snapshot restore, and rollback paths while that lock is held
- add a process-level regression test proving competing environment mutations wait and then continue

## Verification
- AWS Crabbox: `cargo test --locked`
- AWS Crabbox: `cargo check --locked --all-targets`
- AWS Crabbox: `cargo check --locked --target x86_64-pc-windows-gnu`
- AWS Crabbox: all 42 upgrade integration tests, including the new concurrent-mutation test
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -1620,6 +1620,7 @@ impl Cli {
         options: UpgradeOptions,
     ) -> Result<UpgradeEnvSummary, String> {
         let _transaction_lock = lock_upgrade_transaction(name, &self.env, &self.cwd)?;
+        let _operation_lock = self.environment_service().lock_operation(name)?;
         self.upgrade_env_locked(name, target, options)
     }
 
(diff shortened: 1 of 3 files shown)


shakkernerd:
Merged using rebase.

What changed after review:
- Held the per-environment operation lock for the complete normal upgrade lifecycle.
- Used non-relocking binding, service, snapshot restore, and rollback paths while the lock is held.
- Added process-level coverage for a competing environment mutation waiting and then continuing.

Verification:
- AWS Crabbox cbx_a55a7013c2d7: full `cargo test --locked`, all-target checks, and Windows GNU compile passed.
- All 42 upgrade integration tests passed, including the concurrent-mutation regression.
- Auto-review: lock ordering, nested mutation paths, rollback recovery, and sibling environment/service callers reviewed clean.
- GitHub checks: all required checks green on the prepared head.

Landed on `main`:
- Final landed SHA: 3a76b993e43ae3eee3bcf274057653322cb3baa8; range b623410df0d4ea6d230d54ac6a377a84e8bf5a16...3a76b993e43ae3eee3bcf274057653322cb3baa8.

Thanks @shakkernerd!

PR #64, opened 2026-08-06
fix: omit workspace temp files from archives

fuller-stack-dev:
## Summary

- omit each configured OpenClaw workspace's `.openclaw/tmp` subtree from environment archives
- preserve normal workspace files, ordinary `tmp` directories, `.openclaw/tmp-keep` near misses, and configured secondary workspaces
- cover the archive and restore behavior with an end-to-end regression test

## Motivation

OCM currently treats the complete configured workspace as durable. That explicit include overrides the normal archive filter, so workspace-local OpenClaw temporary checkouts and build artifacts are copied into every snapshot.

In a representative 128 MiB transient fixture, the archive shrank from 135,271,936 bytes to 1,053,184 bytes while retaining the 1 MiB durable workspace file. A measured production-shaped environment had about 13 GB under this exact temporary root.

## Verification

- `cargo fmt --check`
- `cargo test store::openclaw_state::tests:: --lib`
- `cargo test --locked --test env_snapshot_tests`
- `cargo test --locked --test store_flow_tests -- --test-threads=1`
- `cargo test --locked -- --test-threads=1 --skip bin_wrapper_runs_with_an_overridden_home`
- source-blind CLI snapshot and restore contract: 5/5 clauses passed
- autoreview: clean, no accepted or actionable findings

The skipped wrapper test fails unchanged on canonical `main` in this local Homebrew Rust/rustup layout. It does not exercise archive behavior.

### DIFF
--- a/src/store/openclaw_state.rs
+++ b/src/store/openclaw_state.rs
@@ -105,10 +105,16 @@ pub(crate) fn openclaw_env_archive_options(
     runtime: OpenClawWorkspaceRuntime<'_>,
 ) -> Result<EnvArchiveOptions, String> {
     let workspaces = resolve_env_openclaw_workspaces(paths, env, runtime)?;
+    let included_path_roots = workspaces.archive_relative_roots(&paths.root)?;
+    let mut excluded_path_roots = openclaw_archive_excluded_path_roots(paths)?;
+    excluded_path_roots.extend(openclaw_workspace_archive_excluded_path_roots(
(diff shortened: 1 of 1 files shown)


fuller-stack-dev:
Superseded by https://github.com/shakkernerd/ocm/pull/65 at the identical code tree because the active GitHub Actions outage dropped the original workflow triggers.

PR #66, opened 2026-08-09
fix(upgrade): quiesce runtime rollback mutations

fuller-stack-dev:
## Summary

- track runtime mutation ownership per transaction so reused named targets are not backed up, restored, or removed
- quiesce managed services and wait for supervisor acknowledgement before runtime-tree mutation
- restore only owned runtime bytes, restore the snapshot binding before created-runtime cleanup, and reassert the prior running policy

## Scope

This intentionally does not include snapshot excluded-state preservation. PR #60 owns that separate archive/restore invariant and its live head passes `environment_snapshot_restore_preserves_current_excluded_state`.

## Root cause

Upgrade transactions treated every pre-existing candidate runtime as mutated. During a failed switch to an already-installed named runtime, rollback recopied that unchanged target while the supervisor could still retry it from backoff. Mutation ownership was also recorded only after installer success, which could miss partial failures.

## Proof

Base: `c298f1d980b4a4f461f97dc39f9374f5badfb628`

Head: `c06901f4eaee0915e65297c96ebbd8e111463959`

- red on unmodified base: `failed_named_runtime_target_in_backoff_is_not_rewritten_during_rollback` observed the target entrypoint or a chunk disappear
- green on head: failed reused target enters backoff without missing files; target entrypoint/chunk inodes and entrypoint hash remain unchanged
- in-place update control restores original runtime bytes and observes quiescence before both update and rollback mutation
- successful local-launcher to published-runtime switch remains unchanged
- tracked-runtime no-op restores its prior running policy after the installer reports reuse
- focused suites: 44 upgrade, 35 service, and 29 store tests passed serially
- complete `cargo test -- --test-threads=1` passed
- `cargo fmt --check`, `cargo check`, and `git diff --check` passed
- Clippy passed with the base-identical `clippy::large_enum_variant` warning allowed
- source-blind validation passed the failed-target case twice, the in-place rollback case, and the successful-switch control
- autoreview reported no accepted/actionable findings

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -182,7 +182,6 @@ struct UpgradeTransactionPlan {
 struct UpgradeRollbackPlan {
     record: UpgradeHistoryRecord,
     recovery: Option<UpgradeRuntimeRecovery>,
-    service: Option<ServiceSummary>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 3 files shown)


fuller-stack-dev:
Squash-merged as `b80b5c5e48c5c384d14b0a1be7ba6d59cfc68166`. All required CI checks passed against reviewed head `c06901f4eaee0915e65297c96ebbd8e111463959`.

PR #67, opened 2026-08-10
fix: preserve durable OpenClaw state during cleanup

fuller-stack-dev:
## Summary

- preserve OpenClaw's durable `.openclaw/state` directory during nonportable runtime-state cleanup
- continue removing foreign environment session references during snapshot restore and repair
- add unit and integration coverage for the device-pairing SQLite state

## Root cause

OpenClaw now stores paired-device records in `.openclaw/state/openclaw.sqlite`. OCM included that database in environment snapshots, but the post-restore repair path treated the root-level `state` directory as disposable whenever it detected stale or foreign environment references. The archive therefore contained the pairing database, then cleanup deleted it after restore.

## Reproduction and validation

On the unmodified base, restoring the same snapshot with a foreign session reference removed `state/openclaw.sqlite`; a control environment without that reference retained its pairing row.

With this change, the same restore contract preserves the paired-device token while still removing the foreign session reference. A repeated source-blind restore also retained the pairing row, and the locally deployed binary kept Rosita healthy and ready.

## Tests

- `cargo fmt --check`
- `cargo test --test env_snapshot_tests env_snapshot_restore_preserves_device_pairing_state_while_clearing_foreign_runtime_refs -- --exact --nocapture`
- `cargo test --lib -- --test-threads=1` (288 passed)
- autoreview: clean, no accepted/actionable findings

### DIFF
--- a/src/store/openclaw_state.rs
+++ b/src/store/openclaw_state.rs
@@ -187,7 +187,7 @@ pub(crate) fn clear_nonportable_runtime_state(
         let path = entry.path();
         let name = entry.file_name();
         if name == OsStr::new("openclaw.json")
-            || is_managed_plugin_state_root(&name)
+            || is_durable_openclaw_state_root(&name)
             || workspaces.contains(&path)
         {
             continue;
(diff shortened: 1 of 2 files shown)


fuller-stack-dev:
Landed via squash as `8079001acc30cb853fd3302f6724090763fee5d3` after all required checks passed.

PR #70, opened 2026-08-11
test: synchronize progress heartbeat checks

shakkernerd:
## Summary
- replace scheduler-dependent sleeps with heartbeat event synchronization
- keep bounded failure timeouts for missing progress output

## Verification
- `cargo fmt --check`
- 50 focused progress-test runs
- `cargo test --locked --lib`
- `git diff --check`

### DIFF
--- a/src/cli/mod.rs
+++ b/src/cli/mod.rs
@@ -708,7 +708,7 @@ impl Cli {
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


shakkernerd:
Fixed: progress heartbeat tests now synchronize on emitted events and no longer depend on runner scheduling.

With thanks to @shakkernerd.

PR #73, opened 2026-08-11
fix(runtime): support current OpenClaw build-local

fuller-stack-dev:
## Summary

- prefer OpenClaw's current `scripts/ocm-npm-workspace-deps.mts` adapter while retaining `.mjs` fallback compatibility
- compute the complete transitive workspace dependency closure across dependencies, optional dependencies, peer dependencies, and dev dependencies
- handle cycles without repacking the root package, and fail closed when a referenced workspace package is missing
- force the trusted local `npm pack` lifecycle even when ambient npm configuration has `ignore-scripts=true`

Archive rewriting remains owned by OpenClaw. This PR does not add managed-Node support, snapshot/rollback behavior, or convergence retries.

## Review decisions

- omitted the supplied install-command context change because it was unrelated diagnostic polish rather than part of the broken build-local invariant
- corrected the supplied closure implementation by seeding the root package as visited, preventing a dependency cycle back to the root from packing it as a workspace dependency
- kept the implementation at the existing runtime build-local owner in `src/store/runtimes.rs`

This is the focused runtime subset overlapping #60; it intentionally excludes that PR's lifecycle, snapshot, and retry changes.

## Proof

Red reproduction on unmodified `d1c36adda05c1129af9788be35528bc1240d95da` confirmed three independent failures:

1. the `.mts` workspace adapter was not invoked
2. nested workspace dependencies were omitted
3. ambient `ignore-scripts=true` silently skipped trusted prepack

Validation on `c22dc5b41c1b0d12603e1cd80a5285748c02e3c3`:

- focused build-local tests: 6/6 passed (18.88s)
- runtime unit tests: 3/3 passed
- full `cargo test --locked` with `ulimit -n 4096`: passed (122.41s)
- `cargo check --workspace --all-targets --locked`: passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- `cargo build --release --locked`: passed (59.52s)
- source-blind behavior validation: passed (2.31s)
- autoreview: no actionable findings, confidence 0.94; TruffleHog clean

Exact real OpenClaw `build-local` E2E used:

- OCM: `c22dc5b41c1b0d12603e1cd80a5285748c02e3c3`
- OpenClaw base: `5179e1235293bbd80c623cf3b6359414b1cb1e74`
- transitive tarball head: `22ad7d29af7c0da2d50259ab3f0365a7b9820e8f` (openclaw/openclaw#122214)
[...]

PR #74, opened 2026-08-12
fix(upgrade): quiesce rejected bindings before rollback

fuller-stack-dev:
## Summary

- quiesce a rejected active binding before restoring an upgrade snapshot, even
  when the target reused an existing named runtime and mutated no runtime tree
- prevent a backoff/retry process from rewriting managed-plugin
  `node_modules/openclaw` peer links after source-state restoration
- extend the named-runtime rollback regression to cover four managed plugins,
  retained/missing/rejected links, restored service binding, and gateway RPC
  readiness

## Root cause

`rollback_upgrade_locked` quiesced only when runtime bytes had changed. A switch
to an already-installed named runtime changes the published binding without
adding `mutated_runtime_names`, so a rejected target in supervisor backoff could
remain able to mutate the environment while OCM restored the source snapshot.

Whole-root snapshot restoration already preserves these symlinks in a cold
fixture. A post-rollback repair phase would therefore be downstream and still
racy. The owner invariant is process exclusion: quiesce whenever rollback
changes runtime bytes or the active binding, before restoring source-owned
state.

## Evidence

Base: `6a30e9063bcf85d1cba4c1c358ce6140e70fcf40`

Head: `f63f41af1638a3c03a6f6826d908b128f89d13ce`

On unmodified canonical source, the active-service fixture completed rollback
with `clickclack/node_modules/openclaw` resolving to rejected runtime
`2026.8.2` instead of restored runtime `2026.8.1`.

On this head, the same fixture reports:

```text
rollback proof: binding=2026.8.1 plugins=4 peerTarget=.../runtimes/2026.8.1/files/node_modules/openclaw gatewayRpc=ok channels=fixture-not-supported
```

The fixture verifies `clickclack`, `codex`, `discord`, and `llama-cpp`; exact
raw and resolved peer-link targets; one retained valid link; one removed link;
two rejected-runtime links; restored service binding; and deep gateway RPC
readiness. It does not configure real channels, so no channel-health claim is
made.

## Validation

- `cargo test --locked -- --test-threads=1`
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings -A clippy::question_mark -A clippy::large_enum_variant -A clippy::too_many_arguments`
- `cargo +1.88.0 check --workspace --all-targets --locked`
- source-blind red/base and green/head behavior validation
- focused cold-snapshot falsifier and rollback readiness/service controls
- shared autoreview: clean, no actionable findings

PRs #71, #72, and #73 and current `main` were checked before implementation.
This change does not add OpenClaw convergence retries or invoke a manual repair
command.

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -3391,11 +3391,13 @@ impl Cli {
         transaction: &UpgradeTransaction,
     ) -> Result<(), String> {
         let changes_runtime_trees = !transaction.mutated_runtime_names.is_empty();
-        if changes_runtime_trees {
+        let changes_binding = transaction.source.kind != transaction.target.kind
+            || transaction.source.name != transaction.target.name;
+        if changes_runtime_trees || changes_binding {
             self.service_service()
                 .quiesce_for_runtime_mutation_locked(env_name)
                 .map_err(|error| {
-                    format!("failed to quiesce the failed target before runtime rollback: {error}")
+                    format!("failed to quiesce the failed target before upgrade rollback: {error}")
                 })?;
         }
         for runtime_backup in transaction.runtime_backups.iter().filter(|backup| {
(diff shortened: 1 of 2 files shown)


fuller-stack-dev:
Landed via squash merge as `f6dea9f7e90215964d8a5dfc25c841c71f298629` after all required checks passed.

PR #80, opened 2026-08-15
fix(upgrade): validate managed Codex candidate before cutover

fuller-stack-dev:
## What problem this solves

An OCM upgrade could pass gateway readiness and enter the transaction even when the staged OpenClaw runtime's managed Codex app-server was missing or the wrong version. Gateway readiness does not start Codex, so the first post-cutover Codex turn could fail after source quiescence and runtime mutation had already begun.

## Canonical fix

OCM now runs the staged candidate binary's named read-only health check before creating the upgrade transaction:

`doctor --lint --only codex/managed-app-server --json`

The check runs through the candidate's normal launch policy and environment using the exact prepared binary path. A real check failure exits through the existing pre-cutover failure path, leaving source binding, service desired/running state, configuration, snapshots, and history unchanged. Older OpenClaw candidates that do not expose the named check retain backward-compatible upgrade behavior through narrowly matched unknown-check/unsupported-doctor responses.

OpenClaw PR https://github.com/openclaw/openclaw/pull/124137 owns the named managed-binary/version check. OpenClaw PR #124124 already owns exact plugin-root binary selection. OCM owns only candidate invocation and transaction ordering; neither OpenClaw concern is duplicated here.

## Evidence

- Red on detached canonical OCM main `0686a2b7003403d092be0b7f3ddcf3f02a6413b6`: the staged target proceeded to config validation/finalization without any managed-Codex candidate check.
- Green: all 48 `upgrade_command_tests` pass, including exact prepared-binary invocation, pass/fail/unsupported compatibility, pre-transaction ordering, staged official installs, and negative controls for lookalike errors.
- Full unit suite reached 298/298 passing unit tests; the later `bin_wrapper_tests` failure reproduces unchanged on the exact unmodified base and is unrelated environment baseline noise.
- `cargo clippy --all-targets --all-features` passes with the same four pre-existing warnings as main. `cargo fmt --all -- --check` and `git diff --check` pass.
- Source-blind failure proof: a candidate reporting Codex 0.146.0 instead of required 0.147.0 exits 1 before cutover. Source runtime binding remains `old-local`, service remains disabled/stopped, snapshots remain `[]`, no upgrade history is created, and config SHA-256 remains unchanged.
- Source-blind success proof: exact candidate order is `--version`, managed-Codex doctor check, `config validate`, `update finalize`, `--version`; outcome switches to `new-local` and retains one `pre-upgrade` snapshot recording `old-local`.
- An unrelated environment remains untouched, with no command log or service-state change.
- Autoreview used the trusted TruffleHog binary and returned clean with no accepted/actionable findings.

This PR is separate from the already-landed crash-loop convergence repair in #79 and the OpenClaw doctor schema-safety repair in openclaw/openclaw#124110.

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -13,6 +13,7 @@ use serde_json::{Value, json};
 use super::{Cli, render};
 use crate::env::{
     CloneEnvironmentOptions, CreateEnvSnapshotOptions, EnvDevMeta, RestoreEnvSnapshotOptions,
+    resolve_runtime_run_dir,
 };
 use crate::infra::shell::{build_openclaw_dev_source_env, build_openclaw_env};
 use crate::openclaw_repo::{detect_openclaw_checkout, ensure_openclaw_worktree};
@@ -22,7 +23,7 @@ use crate::runtime::releases::{
 };
 use crate::runtime::{
     InstallRuntimeFromOfficialReleaseOptions, OfficialRuntimePrepareAction, RuntimeMeta,
-    RuntimeReleaseSelectorKind, RuntimeService, StagedRuntimeInstall,
+    RuntimeReleaseSelectorKind, RuntimeService, StagedRuntimeInstall, resolve_runtime_launch,
 };
 use crate::service::ServiceSummary;
 use crate::store::{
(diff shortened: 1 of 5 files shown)


fuller-stack-dev:
Updated exact head `41138a5c6464c6631605245859ae848df3144efb` after the first CI run exposed an adjacent availability fixture gap.

The production code correctly rejected that fake runtime's generic `unexpected command`; the fixture now models an older OpenClaw candidate with the exact unknown-check JSON compatibility contract. Production matching remains unchanged and narrowly rejects lookalike failures. The previously failing availability test now passes with its stop-to-ready timing proof, all 48 upgrade command tests remain green, formatting/diff checks pass, and the fixture-only delta received a clean autoreview/TruffleHog result at 0.99 confidence.

PR #83, opened 2026-08-18
fix(migrate): relocate imported plugin records

Patrick-Erichsen:
## Summary

- relocate managed ClawHub and npm plugin install records into the imported environment state root during `ocm adopt import`
- handle copied fixtures whose records still reference the original gateway state root without retaining source references
- require canonically contained copied payloads, rejecting traversal and symlink escapes
- reject non-regular or escaping imported SQLite paths before opening the plugin index for write
- preserve external development checkout records and emit an explicit isolation warning instead of copying or mutating arbitrary external paths

## Validation

- `cargo test --test migrate_command_tests` (40 passed)
- `cargo test --lib` (315 passed)
- `cargo fmt --check`
- `cargo clippy --all-targets --all-features` (passed with four warnings)
- structured autoreview against `origin/main` (no actionable findings)

The symlink regression fails on the original PR head because the import succeeds and follows the copied database symlink. With the guard, the import rejects the path, removes the partial environment, and leaves the external SQLite database byte-for-byte unchanged.

## Behavior proof

The rebased binary imported a task-owned plain OpenClaw home with a managed plugin and an external workspace into an isolated `OCM_HOME`. No launcher or service was created.

```console
$ OCM_HOME=<tmp>/ocm-home PATH=/usr/bin:/bin target/debug/ocm adopt import --name proof <tmp>/source/.openclaw --json
{
  "name": "proof",
  "sourceName": "plain-openclaw",
  "root": "<tmp>/ocm-home/envs/proof",
  "archivePath": "<tmp>/source/.openclaw",
  "defaultRuntime": null,
  "defaultLauncher": null,
  "protected": false
}

$ sqlite3 <tmp>/ocm-home/envs/proof/.openclaw/state/openclaw.sqlite \
    "SELECT install_records_json FROM installed_plugin_index WHERE index_key='installed-plugin-index';"
{"demo":{"source":"clawhub","installPath":"<tmp>/ocm-home/envs/proof/.openclaw/extensions/demo"}}

$ jq -r '.agents.defaults.workspace' <tmp>/ocm-home/envs/proof/.openclaw/openclaw.json
<tmp>/ocm-home/envs/proof/.openclaw/workspaces/default
```

`cmp` confirmed that the copied workspace file matches the external source byte-for-byte. `ocm env list --json` reported `serviceEnabled: false` and `serviceRunning: false` for the proof environment.

### DIFF
--- a/src/cli/migrate.rs
+++ b/src/cli/migrate.rs
@@ -168,6 +168,7 @@ impl Cli {
             result.cleared_sandbox_origin,
             result.sandbox_port,
         );
+        self.warn_external_migration_plugins(&result.external_plugin_ids);
         let summary = result.summary;
 
         if json_flag {
@@ -209,6 +210,7 @@ impl Cli {
             result.cleared_sandbox_origin,
             result.sandbox_port,
         );
+        self.warn_external_migration_plugins(&result.external_plugin_ids);
         let summary = result.summary;
 
         if json_flag {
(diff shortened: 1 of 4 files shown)


shakkernerd:
@clawsweeper review

jalehman:
@clawsweeper re-review

jalehman:
@clawsweeper re-review

PR #87, opened 2026-08-18
fix: reject cross-checkout dependency trees

jalehman:
## What

Reject OpenClaw source workflows when the selected checkout’s top-level `node_modules` resolves through another checkout. The validation runs before local package construction and before source watch stops a managed service.

## Why

A linked worktree can point `node_modules` at a different checkout. OCM then combines source from the selected commit with workspace packages from another commit, which can fail during plugin loading or run incompatible code.

## Changes

- Validate checkout-owned dependency resolution
- Reject dangling dependency symlinks
- Block takeover before service stop
- Report isolated-checkout remediation
- Cover build and takeover regressions

## Maintainer decision

OCM should reject cross-checkout dependency trees. Combining source from one checkout with workspace packages from another does not represent either revision and caused the primary Gateway test failure that motivated this change. The error preserves the checkout and gives an isolated-checkout remediation.

## Real behavior proof

Ran the PR-head OCM binary against the real `.worktrees/plugins-hub-stable-shell` linked worktree, whose `node_modules` resolves to the root OpenClaw checkout. An isolated temporary `OCM_HOME` kept the primary environment out of scope.

```text
$ OCM_HOME=<temporary-home> cargo run --quiet -- runtime build-local pr87-real-worktree-proof --repo <openclaw>/.worktrees/plugins-hub-stable-shell --raw
ocm: OpenClaw dependencies resolve outside the selected checkout: <openclaw>/.worktrees/plugins-hub-stable-shell/node_modules -> <openclaw>/node_modules. OCM will not build or run source against dependencies from another checkout. Use a standalone checkout at the selected commit, run `pnpm install --frozen-lockfile` there, and pass that checkout with `--repo`.
Run "ocm help" for usage.
proof_exit=1

$ find <temporary-home>/runtimes -mindepth 1 -maxdepth 1 -print
<no output>
```

The rejection occurred before `npm pack`, runtime creation, or service takeover.

## Review follow-up

The first review found that `Path::exists()` treated a dangling `node_modules` symlink as absent. Both affected entry-point tests failed on the previous PR head because package construction and forced takeover proceeded:

```text
runtime_build_local_rejects_dangling_dependency_link_before_pack ... FAILED
dev_watch_force_rejects_dangling_dependency_link_before_takeover ... FAILED
```

The shared dependency-owner check now uses symlink metadata to distinguish an absent path from a dangling link, then canonicalizes every present path. Both tests pass and verify that `npm pack` does not run and the managed service remains running.

## Testing

- `cargo test --locked --test runtime_command_tests -- --test-threads=1` (48 passed)
- `cargo test --locked --test dev_command_tests -- --test-threads=1` (41 passed)
- Structured autoreview: no actionable findings
- Real linked-worktree build: rejected before pack

The runtime suite’s existing concurrent-install test flaked during earlier full-suite runs and passed on its exact rerun. `bin_wrapper_runs_with_an_overridden_home` also fails on this host because its isolated `HOME` cannot initialize Rustup; the focused changed-surface suites pass.

### DIFF
--- a/src/cli/dev.rs
+++ b/src/cli/dev.rs
@@ -32,7 +32,8 @@ use crate::infra::process::run_direct;
 use crate::infra::shell::{build_openclaw_dev_source_env, build_openclaw_env};
 use crate::infra::terminal::{Cell, KeyValueRow, Tone, paint, render_key_value_card, render_table};
 use crate::openclaw_repo::{
-    detect_openclaw_checkout, discover_openclaw_checkout, ensure_openclaw_worktree,
+    detect_openclaw_checkout, discover_openclaw_checkout, ensure_checkout_owned_dependencies,
+    ensure_openclaw_worktree,
 };
 use crate::service::service_backend_support_error;
 use crate::store::{
@@ -411,6 +412,7 @@ impl Cli {
                 display_path(&repo_root)
             )
         })?;
+        ensure_checkout_owned_dependencies(&repo_root)?;
         let meta = self
             .environment_service()
             .apply_effective_gateway_port(existing)?;
@@ -704,6 +706,7 @@ impl Cli {
             .as_ref()
             .ok_or_else(|| format!("environment \"{}\" is missing its dev binding", meta.name))?;
         let worktree_root = Path::new(&dev.worktree_root);
+        ensure_checkout_owned_dependencies(worktree_root)?;
         let pnpm_store = worktree_root.join("node_modules").join(".pnpm");
         let tsx_bin = worktree_root.join("node_modules").join(".bin").join("tsx");
         if pnpm_store.exists() && tsx_bin.exists() {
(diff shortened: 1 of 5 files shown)


jalehman:
@clawsweeper re-review

The PR body now records the maintainer decision to require checkout-owned dependencies and includes a redacted transcript from the real linked worktree showing rejection before runtime creation.

jalehman:
Maintainer decision: accept strict checkout isolation. A release-shaped runtime build must not inherit dependencies from another checkout. Existing linked worktrees with cross-checkout `node_modules` should fail before runtime creation; operators can install dependencies inside the selected checkout or use an explicitly source-backed development environment.

jalehman:
@clawsweeper re-review

jalehman:
@clawsweeper re-review

jalehman:
Clarification: I maintain OpenClaw and operate the affected OCM environment, but this account does not have push access to `openclaw/ocm`. The decision above records OpenClaw consumer and operator sponsorship. An OCM repository maintainer still needs to accept the strict compatibility policy.

PR #92, opened 2026-08-18
fix: timeout and cap runtime and self-update downloads

SebTardif:
## What Problem This Solves

`ocm self update` and runtime URL installs use ureq 3 with no connect or global timeout, then `io::copy` the body with no size cap. A stalled or huge URL can hang the CLI or fill the disk.

## Why This Change Was Made

ureq 3 defaults leave `timeout_connect` and `timeout_global` unset. Downloads need both a time bound and a byte cap.

## User Impact

A hung GitHub or runtime host fails after 30s. A body over 512 MiB is rejected.

## Evidence

terminal output from the capped copier:

```console
$ cargo test --lib copy_capped -- --nocapture
test infra::download::tests::copy_capped_errors_when_reader_exceeds_max ... ok
test infra::download::tests::copy_capped_copies_up_to_max ... ok
```

## Real behavior proof

Behavior addressed: Runtime and self-update downloads time out at 30s and fail closed above 512 MiB.
Real environment tested: macOS, Rust toolchain, clone at /tmp/pr-ocm on the patched branch.
Exact steps or command run after this patch: cargo test --lib copy_capped
Evidence after fix: terminal output copied below.

```console
$ cargo test --lib copy_capped
ok. 3 passed
```

Observed result after fix: A reader that exceeds the cap returns an error. Exact-size reads still succeed.
What was not tested: A live hanging HTTPS server during `ocm self update`.

### DIFF
--- a/src/cli/self_cmd.rs
+++ b/src/cli/self_cmd.rs
@@ -7,7 +7,7 @@ use semver::Version;
 use serde::Serialize;
 
 use crate::infra::archive::extract_tar_gz;
-use crate::infra::download::{download_to_file, verify_file_sha256};
+use crate::infra::download::{download_to_file, http_agent, verify_file_sha256};
 use crate::store::resolve_ocm_home;
 
 use super::{Cli, render};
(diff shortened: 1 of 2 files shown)


shakkernerd:
Fixed: Runtime and self-update downloads now fail on stalled network phases and oversized artifacts without rejecting slow progressing transfers.

With thanks to @SebTardif.

PR #95, opened 2026-08-19
test: stabilize upgrade availability health probe

shakkernerd:
## What

Treat transport-level misses from the controlled health fixture as inconclusive samples while preserving explicit HTTP 200 and 503 observations.

## Why

The macOS main run recorded a false outage when one fixture response exceeded the poller's transport timeout. The fixture remained available, and the identical tree passed macOS before merging.

## Testing

- `cargo fmt --check`
- Targeted upgrade availability test passed 10 consecutive times
- Temporary pre-cutover HTTP 503 injection triggered the existing outage assertion as expected

### DIFF
--- a/tests/upgrade_availability_tests.rs
+++ b/tests/upgrade_availability_tests.rs
@@ -209,22 +209,29 @@ impl ControlledHealthServer {
     }
 }
 
-fn health_ok(port: u32) -> bool {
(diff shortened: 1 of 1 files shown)


shakkernerd:
Fixed: The upgrade availability proof now ignores inconclusive transport misses while still detecting explicit gateway outages.

PR #103, opened 2026-08-21
fix: harden gateway restart, cleanup, and child temp dirs

hannesrudolph:
Related: #60

<details>

</details>

## What Problem This Solves

`ocm service restart` directly replaces a supervised gateway even when OpenClaw advertises restart-handoff support, so eligible active sessions and subagents can be interrupted without receiving OpenClaw's recovery marker. A restart issued from inside the target gateway can also wait on its own replacement.

Environment removal and service uninstall currently reconcile the complete supervisor registry. If sibling runtime metadata or inherited process environment has drifted, deleting one environment can rewrite or restart unrelated gateways. Exited children can leave descendants or an unreaped process-group leader behind, and persisted macOS per-login `TMPDIR` values can become invalid before a later gateway spawn.

## Why This Change Was Made

Use OpenClaw restart-handoff protocol v1 for normal service restarts, return immediately for a self-restart, retain a warned direct-restart fallback for legacy bindings, and expose `--force` as the explicit recovery bypass.

Make uninstall, guarded destroy, direct removal, and prune reconcile only the affected environment. Clean the complete child process group with TERM, bounded polling, KILL fallback, and an unconditional leader wait. Stop persisting `TMPDIR`; each child spawn receives a private daemon-scoped mode-`0700` directory derived from the daemon's current temp base, with a safe fallback when that base is unavailable.

Local companion packaging and named Tailscale Service ingress changes are intentionally outside this PR.

## User Impact

Gateway restarts preserve OpenClaw's eligible recovery behavior instead of blindly interrupting active work. Removing or stopping one environment leaves sibling gateway definitions and PIDs unchanged despite unrelated drift. Gateway descendants are cleaned up reliably, and future child spawns no longer inherit stale temporary-directory paths after login, reboot, or daemon replacement.

Legacy gateways remain restartable through the existing direct supervisor path and receive an explicit warning; operators can deliberately select that path with `--force`.

## Evidence

- Isolated split branch based directly on current `main`; it contains no companion-packaging or Tailscale ingress changes.
- `cargo fmt --all -- --check`: passed.
- `git diff --check origin/main..HEAD`: passed.
- `cargo test --locked --test daemon_runtime_tests --test env_destroy_tests --test service_command_tests --test upgrade_command_tests`: 126 passed, 0 failed.
- Hosted CI passed on macOS, Ubuntu, Windows, formatting, and the Rust 1.88 minimum lane.
- The original umbrella head also passed 183 focused tests and its complete hosted CI matrix after rebasing.
- No OCM binary was installed locally, and no resident daemon or managed gateway was restarted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ All notable changes to OCM are documented here.
 - Copy repo-backed and symlinked plain-home agent workspaces into an adopted
   environment and rewrite the imported config to keep the fixture isolated.
 
+### Changed
+
+- Make `ocm service restart <env>` restart immediately through OpenClaw's protocol-v1 recovery handoff so eligible interrupted sessions and subagents resume after startup, preserve the legacy direct-restart behavior with a warning when recovery is unavailable, keep `--force` as an explicit bypass for an unhealthy handoff, and avoid self-restart deadlocks.
+
 ## 0.2.30 - 2026-07-23
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to OCM are documented here.
 ### Fixed
 
 - Keep unrelated supervised gateways on their persisted child specifications when an environment is destroyed, removed, or pruned, so latent metadata or process-environment drift cannot restart sibling gateways during cleanup.
+- Resolve `TMPDIR` from the live OCM daemon whenever a gateway is spawned, create a daemon-scoped private temporary directory, and fall back to a private directory under `/tmp` instead of persisting a per-login path across reboots.
 
 ## 0.2.30 - 2026-07-23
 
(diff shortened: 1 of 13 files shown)


hannesrudolph:
@clawsweeper re-review

I accepted the activated-shell finding and pushed `7f924cb`.

### Change selected

The self-restart exemption now requires both:

- `OCM_ACTIVE_ENV` matching the target environment; and
- `OPENCLAW_SERVICE_KIND=gateway`.

I used the existing service-kind marker instead of adding another OCM-only variable. `OCM_ACTIVE_ENV` is deliberately exported by `ocm use`, while `OPENCLAW_SERVICE_KIND=gateway` is applied only to a protocol-v1 supervised gateway child. Activated shells therefore continue through replacement observation and readiness verification, while a real gateway-originated restart still returns immediately and avoids waiting on itself.

### Coverage

- The true self-restart fixture now carries the gateway service-kind marker and still proves the no-wait handoff path.
- The readiness integration test now performs restart with only `OCM_ACTIVE_ENV` set, reproducing an activated shell and requiring `gatewayReady: true` after replacement.
- Full focused branch validation: 126 passed, 0 failed across daemon runtime, environment destroy, service command, and upgrade command suites.
- Formatting and `git diff --check` passed.

I did not decline any part of the actionable review advice; the only implementation choice was to reuse the existing protocol-v1 gateway identity instead of creating a new marker.

PR #105, opened 2026-08-21
fix(env): destroy works from wrappers that mention the target path

### LINKED ISSUE #104: env destroy can count its own wrapper and inspection processes
## Summary

`ocm env destroy` can classify its own caller wrapper and transient inspection
children as processes owned by the target environment. The guarded destroy then
fails with:

```text
environment process state changed during inspection; retry the command
```

This occurs even when the target service is stopped, `childPid` is null, and
there are no open files under the environment root.

## Reproduction

1. Create or clone a disposable environment and stop its service.
2. Invoke `ocm env destroy <env> --json` from a `zsh -c` wrapper whose command
   line contains the target environment root, for example because the same
   wrapper also runs `lsof`, `test`, or another target-path check.
3. Observe the preview or guarded apply fail because the two process-table
   scans disagree.
4. Replace the wrapper with a direct `exec` whose process command line does not
   contain the target root.
5. The same stopped environment reports `processCount: 0`, and guarded destroy
   succeeds.

Observed on macOS with OCM 0.2.32. The target had no live child PID and no open
files. Production sibling PIDs remained unchanged throughout.

## Cause

`associated_processes_once` seeds environment ownership when
`process_belongs_to_env` finds any environment path in a process command line:

```rust
if markers
    .iter()
    .any(|marker| command_contains_process_path(&command, marker))
{
    return true;
}
```

`interactive_shell_command` excludes only bare shell command strings such as
`zsh` or `/bin/zsh`. A normal wrapper appears as `/bin/zsh -c ...`, so it is not
excluded. Its command text can contain the target root, and transient child
commands can make the two consecutive ownership scans differ.

Relevant code:

- `src/cli/env.rs`: `associated_processes`
- `src/cli/env.rs`: `associated_processes_once`
- `src/cli/env.rs`: `process_belongs_to_env`
- `src/cli/env.rs`: `interactive_shell_command`

## Expected behavior

The destroy process fingerprint should represent target-owned processes only.
The OCM caller, its wrapper/ancestors, and inspection descendants must not
become target candidates merely because their command text mentions the target
path.

Possible approaches:

- Exclude the current OCM process and its caller ancestry/inspection
  descendants from candidate discovery.
- Avoid treating arbitrary wrapper command text as ownership when the process
  is part of the active OCM invocation.
- Prefer stronger ownership evidence such as working directory, environment
  identity, supervisor records, or a known target process tree.

## Regression coverage

Add an end-to-end `env destroy` test that:

1. Creates a stopped environment with zero target processes.
2. Runs preview and guarded apply from a wrapper whose command line contains
   the target root and has a short-lived inspection child.
3. Requires stable `processCount: 0`.
4. Requires guarded destroy to succeed.
5. Verifies sibling environment processes remain unchanged.

hannesrudolph:
Closes #104

<details>

</details>

## What Problem This Solves

`ocm env destroy` can classify its own caller wrapper as a process owned by the target environment when the wrapper command line mentions the environment root. Descendant expansion then pulls the active OCM invocation and transient inspection processes into the candidate set, so the two guarded process scans can disagree and fail with:

```text
environment process state changed during inspection; retry the command
```

This blocks preview and guarded destroy even when the target service is stopped and no target-owned process is running.

## Why This Change Was Made

Exclude the active OCM invocation from environment ownership discovery: the current OCM process, its caller ancestry, and descendants spawned by that OCM process are ignored as candidate seeds and during descendant expansion.

The exclusion deliberately does not remove siblings or all descendants of the caller wrapper, and it does not weaken path-based ownership matching. Independent processes whose command line or working directory belongs to the target environment remain detectable and terminable. This keeps the repair scoped to the invocation tree that cannot safely be treated as destroyable target work.

## User Impact

Operators can preview and perform guarded environment destruction from shell wrappers, maintenance scripts, and diagnostic flows whose command lines mention the target path. The destroy guard remains stable instead of counting its own invocation.

Genuinely target-owned processes are still discovered, while independent sibling-environment processes remain untouched.

## Evidence

- The new wrapper regression failed on unchanged `main` with the exact reported inspection-state error.
- After the fix, the same wrapper preview reports `processCount: 0`, guarded destroy succeeds, and an independent sibling process remains running.
- `cargo test --locked invocation_processes_include_ancestors_and_current_descendants_only -- --nocapture`: passed.
- `cargo test --locked --test env_destroy_tests`: 14 passed, 0 failed.
- `cargo fmt --all -- --check`: passed.
- `git diff --check origin/main..HEAD`: passed.
- Hosted CI passed on macOS, Ubuntu, Windows, formatting, and the Rust 1.88 minimum lane.
- No OCM binary was installed locally, and no resident daemon or managed gateway was restarted.

### DIFF
--- a/src/cli/env.rs
+++ b/src/cli/env.rs
@@ -1708,6 +1708,40 @@ fn associated_processes(meta: &EnvMeta) -> Result<Vec<EnvDestroyProcessIdentity>
 }
 
 #[cfg(unix)]
+fn active_invocation_processes(
+    current_pid: u32,
+    parent_map: &std::collections::HashMap<u32, u32>,
+    children_map: &std::collections::HashMap<u32, Vec<u32>>,
+) -> BTreeSet<u32> {
+    // The destroy command and its caller/inspection tree can mention the target
+    // path without being owned by that environment.
+    let mut excluded = BTreeSet::new();
+    let mut cursor = current_pid;
+    while cursor > 1 && excluded.insert(cursor) {
+        let Some(parent) = parent_map.get(&cursor).copied() else {
+            break;
+        };
+        if parent == cursor {
+            break;
+        }
+        cursor = parent;
+    }
+
+    let mut queue = vec![current_pid];
+    while let Some(pid) = queue.pop() {
+        let Some(children) = children_map.get(&pid) else {
+            continue;
+        };
+        for child in children {
+            if excluded.insert(*child) {
+                queue.push(*child);
+            }
+        }
+    }
+    excluded
+}
+
+#[cfg(unix)]
 fn associated_processes_once(meta: &EnvMeta) -> Result<Vec<EnvDestroyProcessIdentity>, String> {
     let processes = process_table()?;
     let cwd_map = process_cwd_map(&processes)?;
@@ -1722,10 +1756,14 @@ fn associated_processes_once(meta: &EnvMeta) -> Result<Vec<EnvDestroyProcessIden
             .or_default()
             .push(process.pid);
     }
+    let invocation_processes =
+        active_invocation_processes(std::process::id(), &parent_map, &children_map);
 
     let mut seeds = BTreeSet::new();
     for process in &processes {
-        if interactive_shell_command(&process.command) {
+        if invocation_processes.contains(&process.pid)
+            || interactive_shell_command(&process.command)
+        {
             continue;
         }
         if process_belongs_to_env(process.pid, &process.command, meta, &cwd_map) {
@@ -1745,7 +1783,9 @@ fn associated_processes_once(meta: &EnvMeta) -> Result<Vec<EnvDestroyProcessIden
                 let Some(process) = processes.iter().find(|process| process.pid == *child) else {
                     continue;
                 };
-                if interactive_shell_command(&process.command) {
+                if invocation_processes.contains(child)
+                    || interactive_shell_command(&process.command)
+                {
                     continue;
                 }
                 if related.insert(*child) {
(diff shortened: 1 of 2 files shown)


hannesrudolph:
@clawsweeper re-review

This PR implements the focused repair proposed on #104.

### Change selected

Process discovery now excludes:

- the active OCM process;
- its caller ancestry; and
- descendants spawned by that OCM process, including transient inspection helpers.

The exclusion applies both when selecting environment-owned seed processes and when expanding their descendants.

### Deliberate boundary

I did not exclude the caller wrapper's entire descendant tree, and I did not weaken command-line or working-directory path ownership. A blanket wrapper-tree exclusion could hide a genuinely target-owned process launched independently by the same maintenance shell. The helper therefore excludes current-process descendants, not siblings, and the unit test proves that distinction.

### Evidence

- The wrapper regression failed on unchanged `main` with `environment process state changed during inspection; retry the command`.
- With the repair, the same wrapper reports `processCount: 0` and guarded destroy succeeds.
- The regression keeps an independent sibling process alive through destroy.
- Invocation-tree unit coverage passes.
- Full `env_destroy_tests`: 14 passed, 0 failed.
- Formatting and `git diff --check` passed.

No local OCM deployment or gateway restart was performed.

hannesrudolph:
@clawsweeper

The platform checks requested by the latest review are now complete:

- macOS: passed
- Ubuntu: passed
- Windows compile: passed
- Rust 1.88 minimum: passed
- Format: passed

I made no additional code change because the review identified no actionable correctness finding. The remaining merge-risk note is already covered by the checked-in wrapper regression: it proves the invocation tree is excluded while an independent sibling process remains detectable and survives destroy.

The review's live-verification command did not reach the Rust test because its setup attempted a Corepack download. That setup failure does not contradict the focused local before/after reproduction or the now-green hosted platform runs. The PR body has been updated with the completed CI evidence.

PR #106, opened 2026-08-22
fix(supervisor): preserve siblings during targeted lifecycle

hannesrudolph:
## Summary

- preserve unrelated running gateway specs and PIDs during targeted `service start`
- fully reconcile fleet state before activating a stopped daemon
- reconcile only the new environment during clone, simulation-clone, and import
- add regressions for running-daemon, stopped-daemon, and clone isolation

## Incidents

Two related fleet-isolation failures were reproduced:

1. A targeted `ocm service start openclaw` rebuilt complete fleet state and
   relaunched unrelated gateways when latent sibling drift was present.
2. `ocm env clone clawpatrol <canary>` rebuilt complete fleet state and
   reloaded production OpenClaw during supervisor convergence.

Both failures came from target-scoped commands invoking fleet-wide supervisor
synchronization.

## Behavior

- When the supervisor daemon is already running, service start synchronizes
  only the named environment.
- When the daemon is stopped, service start fully reconciles fleet state before
  activation so stale persisted sibling definitions are not launched.
- Clone/import paths synchronize only the newly created environment and do not
  regenerate sibling specs.

## Validation

- focused regressions:
  - `service_start_preserves_running_siblings_despite_unrelated_drift`
  - `service_start_reconciles_siblings_before_activating_stopped_daemon`
  - `env_clone_preserves_unrelated_supervisor_child_specs`
  - `env_import_preserves_unrelated_supervisor_child_specs`
- `cargo test --test daemon_runtime_tests -- --test-threads=1`
- `cargo test --all-targets --locked -- --test-threads=1 --skip bin_wrapper_runs_with_an_overridden_home`
- `cargo build --release --locked`

The skipped wrapper test requires `rustup`, which is not installed on the
deployment host. Every executed test passed.

### DIFF
--- a/src/env/lifecycle.rs
+++ b/src/env/lifecycle.rs
@@ -234,7 +234,7 @@ impl<'a> EnvironmentService<'a> {
 
     pub fn clone(&self, options: CloneEnvironmentOptions) -> Result<EnvMeta, String> {
         let meta = clone_environment(options, self.env, self.cwd)?;
-        sync_supervisor_if_present(self.env, self.cwd)?;
+        sync_supervisor_env_if_present(self.env, self.cwd, &meta.name)?;
         Ok(meta)
     }
 
@@ -243,7 +243,7 @@ impl<'a> EnvironmentService<'a> {
         options: CloneEnvironmentOptions,
     ) -> Result<EnvMeta, String> {
         let meta = clone_environment_for_simulation(options, self.env, self.cwd)?;
-        sync_supervisor_if_present(self.env, self.cwd)?;
+        sync_supervisor_env_if_present(self.env, self.cwd, &meta.name)?;
         Ok(meta)
     }
 
@@ -258,7 +258,7 @@ impl<'a> EnvironmentService<'a> {
             self.env,
             self.cwd,
         )?;
-        sync_supervisor_if_present(self.env, self.cwd)?;
+        sync_supervisor_env_if_present(self.env, self.cwd, &result.meta.name)?;
         Ok(result)
     }
 
--- a/src/env/lifecycle.rs
+++ b/src/env/lifecycle.rs
@@ -228,7 +228,7 @@ impl<'a> EnvironmentService<'a> {
 
     pub fn create(&self, options: CreateEnvironmentOptions) -> Result<EnvMeta, String> {
         let meta = create_environment_with_validated_runtime(options, self.env, self.cwd)?;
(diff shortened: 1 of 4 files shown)


hannesrudolph:
@clawsweeper re-review

Addressed the stopped-daemon finding and added the clone-isolation fix exposed
by the August 23 live incident. Running-daemon, stopped-daemon, and clone
regressions pass, as does the complete deterministic all-target suite.

hannesrudolph:
@clawsweeper re-review

Addressed the import-isolation finding on head `7867050`. Running-daemon,
stopped-daemon, clone, and import regressions pass, as does the complete
deterministic all-target suite.

jalehman:
Extended the creation path at `d8c23090550128aee9c91b1d91b979d2073f560c`.

- `EnvironmentService::create` now uses the target-only supervisor reconciliation already introduced by this PR.
- This covers missing-environment `ocm start`, `env create`, new `dev` environments, and migration/adoption creation through their shared service boundary.
- The incident-shaped regression runs `ocm start pr121799-macos-proof --command 'node openclaw.mjs' --cwd <isolated-test-dir> --port 19079 --no-service --json` with caller `NODE_OPTIONS` drift and asserts that both unrelated child specs and PIDs remain unchanged.

Evidence:

- Pre-fix at `786705082866057b3da0c8840d16cb89a6d2dccf`: the regression fails because both sibling specs absorb caller-environment drift.
- Post-fix: the regression passes and both sibling PIDs remain unchanged.
- Production/test delta: `src/env/lifecycle.rs` +1/-1; `tests/daemon_runtime_tests.rs` +74/-0.
- Passed: `cargo fmt --all -- --check`, `cargo check --workspace --all-targets --locked`, focused regression, 30-test daemon suite excluding the named baseline failure below, full locked all-target suite with the same baseline exclusion and the documented rustup-dependent wrapper exclusion, and `cargo build --release --locked`.
- Independent baseline: `env_destroy_removes_only_target_child_despite_unrelated_drift` fails with `environment process state changed during inspection` at both `7867050` and this head. Strict Clippy is also blocked by the pre-existing `clippy::too_many_arguments` finding at `src/migrate.rs:372`.
- Autoreview ran. Its only finding claimed conflicting source-backed children could both become active; the runtime rejects a pending child when a running child owns the same run directory, and this regression creates a disabled target recorded in `skippedEnvs`, so no patch expansion was made.

No live OCM environment or Gateway was used during proof.

Issue #98, opened 2026-08-20
Adopted environment upgrade mutated the source gateway state database


MertBasar0:
## Summary

During an OpenClaw release-validation run, an OCM-adopted disposable environment was upgraded to an exact beta runtime. The candidate did not remain confined to the adopted environment: it migrated the original source gateway's shared state database from schema 7 to schema 8.

The source gateway could no longer start on its original checkout after the validation attempt and required recovery. The OCM checkpoint available for rollback had been created after the migration, so it did not contain a recent schema-7 copy.

## Environment

- Host: Windows 11 with Ubuntu WSL
- OCM: `0.2.32`
- Source OpenClaw: `2026.8.1` at `f329a9489f32e6472ded5437be971609229c9451`
- Candidate: `v2026.8.1-beta.2` at `8f382a202ff1e15833394b481615dcdda99b04d7`
- Source: a real gateway state directory
- Target: an adopted OCM environment with a distinct environment name, state directory, and port

No credentials or user-specific paths are included here.

## Sequence

1. Inspect and import the real gateway into a disposable environment using the documented `ocm adopt inspect`, `ocm adopt plan`, and `ocm adopt import` flow.
2. Install and verify the exact candidate runtime.
3. The adopted environment had `defaultRuntime=null`, so set its runtime explicitly before continuing.
4. Run the candidate upgrade dry-run, upgrade, and start against the adopted environment.
5. Observe that the isolated candidate reports the exact beta version but fails readiness.
6. Inspect the source gateway after the failed validation run.

## Expected behavior

- All candidate migrations and writes stay within the adopted environment's state directory.
- The source gateway's configuration, databases, and schema metadata remain byte-for-byte unchanged.
- A pre-upgrade checkpoint exists before any candidate migration and can restore the adopted copy independently of the source.

## Actual behavior

- The isolated candidate failed startup three consecutive times with:

  `Gateway failed to start: Cannot read properties of undefined (reading 'values')`

- The original source gateway's state database was migrated from schema 7 to schema 8.
- Its schema metadata recorded `app_version=2026.8.1-beta.2`, even though the source gateway had not been intentionally upgraded.
- The source checkout supported schema 7 and could no longer restart against the migrated database.
- The available OCM checkpoint was created after the migration and therefore did not provide a recent schema-7 recovery copy.
- The exact retained path, link, or environment reference through which the candidate reached the source database has not yet been identified.

[...]

PR #110, opened 2026-08-26
fix(migrate): reject external plugin paths during adoption

MertBasar0:
## What Problem This Solves

`ocm adopt import` can retain absolute `sourcePath` or `installPath` values for path plugins outside the adopted environment. The import currently succeeds with a warning, so a candidate runtime can still receive a writable reference to source-owned state even though the environment is intended to be disposable.

## Why This Change Was Made

Containment is a safety boundary, not a compatibility warning. The import now fails before publishing relocated plugin records when any plugin path cannot be isolated inside the target state root. The existing migration transaction then removes the partially created environment.

## User Impact

Adoption no longer produces a runnable environment with external plugin references. Operators must relocate or remove external path plugins before importing. Managed ClawHub and npm plugin records continue to relocate into the adopted state as before.

### Targeted regression test

```text
$ cargo test --test migrate_command_tests adopt_import_rejects_external_plugin_paths_without_mutating_the_source -- --exact --nocapture
running 1 test
test adopt_import_rejects_external_plugin_paths_without_mutating_the_source ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 39 filtered out
```

### Real CLI behavior with a safe temporary fixture

The fixture contains a path plugin outside the source OpenClaw home. Paths are redacted below. The source database and external plugin marker were hashed before and after the production `ocm` command.

```text
$ ocm adopt import --name proof-external-plugin <fixture>/source/.openclaw --json
ocm: imported OpenClaw plugin path(s) could not be isolated inside the env state: external-dev
Run "ocm help" for usage.

exit: 1
target environment removed: yes
source sqlite unchanged: yes (6a5285e2a14ca9cf4df66734c93ef52404fa218b7a8b2fe2f912e3672ad87bc9)
external plugin unchanged: yes (09d7425c9be6e7eae3c094a5884bba879aef74adb6193f54175898965ed44393)
```

Additional CI-equivalent verification on commit `9695362`:

- `cargo fmt --check` passes.
- `cargo test --locked` passes in a clean WSL-native Linux checkout.
- `cargo +1.88.0 check --workspace --all-targets --locked` passes.
- `git diff --check upstream/main...HEAD` passes.

## Maintainer Decision Requested

Please confirm that fail-closed adoption is the intended policy when a path plugin cannot be contained. This intentionally changes the previous warning-and-continue behavior: operators must relocate or remove the external path plugin before retrying adoption.

Related: #98
Related: openclaw/openclaw#129673

### DIFF
--- a/src/store/openclaw_state.rs
+++ b/src/store/openclaw_state.rs
@@ -402,6 +402,12 @@ fn relocate_installed_plugin_index_paths(
             external_plugin_ids.insert(plugin_id.clone());
         }
     }
+    if !external_plugin_ids.is_empty() {
+        return Err(format!(
+            "imported OpenClaw plugin path(s) could not be isolated inside the env state: {}",
+            external_plugin_ids.into_iter().collect::<Vec<_>>().join(", ")
+        ));
+    }
     if !changed {
         return Ok(MigratedRuntimeStateResult {
             changed: false,
--- a/src/store/openclaw_state.rs
+++ b/src/store/openclaw_state.rs
@@ -405,7 +405,10 @@ fn relocate_installed_plugin_index_paths(
     if !external_plugin_ids.is_empty() {
         return Err(format!(
             "imported OpenClaw plugin path(s) could not be isolated inside the env state: {}",
-            external_plugin_ids.into_iter().collect::<Vec<_>>().join(", ")
+            external_plugin_ids
+                .into_iter()
+                .collect::<Vec<_>>()
+                .join(", ")
         ));
     }
     if !changed {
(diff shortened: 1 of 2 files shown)


MertBasar0:
@clawsweeper re-review

Added the requested after-fix production-command proof from a safe temporary fixture. The targeted regression test passes; the real CLI run exits nonzero, removes the partial target environment, and leaves both the source SQLite database and external plugin marker byte-for-byte unchanged. The PR body now also calls out the remaining maintainer policy decision explicitly.

MertBasar0:
@Patrick-Erichsen The requested real-command proof and CI-equivalent local verification are now in the PR body. The remaining maintainer decision is the compatibility policy: this patch intentionally fails closed when an external path plugin cannot be contained, instead of preserving the current warning-and-continue behavior. Could you confirm that fail-closed adoption is the intended policy?

PR #114, opened 2026-08-28
feat: update selected gateways concurrently

hannesrudolph:
## What Problem This Solves

Operators updating several OpenClaw gateways must currently run complete upgrades serially, even when every environment will use the same already-installed runtime. The repeated snapshot, finalization, restart, and verification windows make fleet rollouts unnecessarily long.

## Why This Change Was Made

Adds an explicit `ocm upgrade batch` mode that coordinates selected environments under one exclusive batch lock. It creates a cold checkpoint for every selected environment before any transition, then performs bounded parallel per-environment upgrades while preserving existing upgrade history and rollback behavior.

This is deliberately opt-in and requires `--accept-fleet-outage`. Environments are always selected explicitly, so a recovery gateway can remain outside the batch.

## User Impact

Operators can update a selected gateway fleet concurrently:

```bash
ocm upgrade batch \
  --runtime origin-main-local \
  --envs clawpatrol,personal,openclaw,main \
  --parallel 4 \
  --failure-policy continue \
  --accept-fleet-outage
```

The normal single-environment and serial `--all` workflows are unchanged.

## Evidence

- 323 library tests passed.
- 33 CLI validation tests passed.
- 32 help tests passed.
- 51 upgrade tests passed, including:
  - dry-run creates no state
  - every selected checkpoint exists before finalizers proceed
  - finalizers overlap at the configured concurrency
  - excluded environments remain unchanged
  - `continue` completes healthy environments and leaves failures stopped
- 33 daemon isolation tests passed.
- Release build passed.
- `cargo clippy --lib --bins -- -D warnings -A clippy::too_many_arguments` passed.
- One existing service fixture race and one availability timing test failed only during full-suite load; both passed in isolation.

### DIFF
--- a/src/cli/render/help.rs
+++ b/src/cli/render/help.rs
@@ -527,6 +527,9 @@ pub fn upgrade_help(cmd: &str) -> String {
             format!(
                 "{cmd} upgrade <env> [--version <version> | --channel <channel> | --runtime <runtime>] [--dry-run] [--no-rollback] [--raw] [--json]"
             ),
+            format!(
+                "{cmd} upgrade batch --runtime <runtime> --envs <env,env,...> [--parallel <count>] [--failure-policy rollback|continue] [--accept-fleet-outage] [--dry-run] [--raw] [--json]"
+            ),
             format!("{cmd} upgrade --all [--dry-run] [--raw] [--json]"),
         ],
         &[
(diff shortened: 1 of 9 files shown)


hannesrudolph:
Live production proof completed on August 28, 2026.

- Batch `1787895609-606237000` selected four gateways with `parallel=4` and excluded the Rescue recovery gateway.
- All four fleet checkpoints existed before the transition phase.
- All four upgrade transactions started concurrently and completed successfully.
- Slowest transaction: about 241 seconds; summed serial duration: about 599 seconds.
- Rescue PID/binding/restart count remained unchanged.
- 159 non-empty SQLite databases passed, all channel probes were clean, browsers stopped, and all five Tailnet gateway URLs returned HTTP 200.
- Claw Patrol Codex + clawtest, Personal Codex + Sol/Luna availability, OpenClaw Codex + Terra, and Main browser/Codex acceptance passed.

Durable batch journal and rollback checkpoints were retained.

PR #130, opened 2026-08-31
fix: upgrades no longer reject migratable legacy config

### LINKED ISSUE #107: Upgrade candidate preflight blocks target config migrations
## Summary

OCM 0.2.33 can reject an otherwise valid OpenClaw upgrade before the target
runtime's documented config migration is allowed to run.

`validate_prepared_upgrade_target` runs the candidate binary as:

```text
doctor --lint --only codex/managed-app-server --json
```

It runs against the source environment's current config. OpenClaw also runs its
mandatory `core/doctor/final-config-validation` contribution, even with the
named `--only` selection. If the target release retired a config key and
provided a `doctor --fix` migration, final validation rejects the old shape.
OCM then exits before the later `repair_target_openclaw_config` step that exists
to apply that migration.

## Reproduction

1. Start with an environment on an OpenClaw runtime that accepts
   `ui.assistant`.
2. Prepare a target runtime containing
   openclaw/openclaw#129067, which intentionally retires `ui.assistant` and
   provides a Doctor repair.
3. Run:

   ```text
   ocm upgrade <env> --runtime <target-runtime> --json
   ```

The upgrade fails with a receipt equivalent to:

```json
{
  "outcome": "failed",
  "serviceAction": null,
  "snapshotId": null,
  "note": "upgrade preparation failed before cutover: candidate managed Codex preflight failed ... core/doctor/final-config-validation ... Unrecognized key: \"assistant\" at ui"
}
```

Applying the target runtime's config repair on a clone makes target validation
pass. After applying the equivalent migration to the source environment, the
same OCM upgrade succeeds.

## Expected behavior

The candidate-specific managed Codex check should still fail closed for a real
managed app-server mismatch, but an expected target config migration should not
make that package preflight unreachable.

OCM should run the candidate check in a context where source-config migration
findings cannot block it, or perform the check after the target config repair in
an isolated/snapshot-backed upgrade context.

## Source evidence

On current `main`, `prepare_isolated_upgrade_target` calls
`validate_prepared_upgrade_target` before the upgrade transaction.
`repair_target_openclaw_config` is only called later from the post-core update
path.

The candidate preflight was added in openclaw/ocm#80. The OpenClaw config
retirement and Doctor migration are in openclaw/openclaw#129067.

## Impact

An automated OCM upgrade can be blocked by exactly the compatibility migration
that OCM's later target-config repair is designed to apply. The source service
is preserved, but operators must diagnose and mutate production config
manually before retrying.

shakkernerd:
Closes #107

## What Problem This Solves

Fixes an issue where `ocm upgrade` rejected migratable legacy OpenClaw config before the target release could run its supported repair.

## Why This Change Was Made

Target config repair now runs inside the snapshot-backed upgrade transaction before managed Codex candidate validation. Candidate validation still completes before OpenClaw finalization, binding publication, or supervisor refresh, and a genuine candidate failure restores the snapshot and original service intent.

## User Impact

Operators can upgrade environments whose current config requires a migration supplied by the target OpenClaw release without manually editing production config first.

## Evidence

- Added a pre-fix regression that fails when candidate validation runs before target config repair.
- Covered explicit runtime, official staged, implicit channel-following, manifest-backed in-place, launcher transition, named runtime, and reused target paths.
- Proved a genuine candidate mismatch causes no finalization or transient binding publication and restores config, state, binding, and service intent.
- Passed cloud macOS ARM64 formatting, all-target checks, all 1,007 tests, and a release build on the exact head.
- Ran the exact cloud-built artifact against the clean Nora `2026.7.1` snapshot. The four issue #107 keys no longer caused the original candidate-preparation failure, and OCM restored the stopped baseline when OpenClaw Doctor reached the separate legacy exec-approvals defect tracked by openclaw/openclaw#133813.

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -2664,8 +2664,32 @@ impl Cli {
                     }
                 }
             } else {
-                transaction.mark_post_update_not_needed();
-                (None, false)
+                let config_repaired = match self.prepare_target_openclaw_update(
+                    env_name,
+                    &prepared.meta,
+                    &mut transaction.timings,
+                ) {
+                    Ok(config_repaired) => config_repaired,
+                    Err(error) => {
+                        transaction.mark_post_update_failed(&error);
+                        return self.rollback_failed_upgrade(
+                            env_name,
+                            "runtime",
+                            previous_binding_name,
+                            "runtime",
+                            prepared.name,
+                            prepared.meta.release_version,
+                            prepared.meta.release_channel,
+                            transaction,
+                            error,
+                        );
+                    }
+                };
+                transaction.mark_target_validation_completed(config_repaired);
+                (
+                    config_repaired.then(|| "OpenClaw config repair completed".to_string()),
+                    false,
+                )
             };
             let publish_started = transaction.timings.start();
             let publish_result = if changed {
@@ -3537,7 +3561,7 @@ impl Cli {
             return Ok(());
         }
         Err(format!(
-            "candidate managed Codex preflight failed: {}. Repair or reinstall the target OpenClaw runtime, then rerun the upgrade",
+            "candidate managed Codex preflight failed: {}. Repair or reinstall the staged OpenClaw runtime, then rerun the upgrade; the source environment was not changed",
             output.failure_summary()
         ))
     }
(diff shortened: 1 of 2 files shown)


shakkernerd:
Fixed: OCM upgrades now repair target-compatible OpenClaw config before candidate validation and restore the snapshot on genuine candidate failure.

PR #135, opened 2026-09-02
fix: managed Gateway upgrades no longer stop midway

fuller-stack-dev:
## What Problem This Solves

Fixes an issue where an OCM upgrade launched by a managed Gateway could be killed by that Gateway's own shutdown. The environment was left stopped on its old runtime, and OCM recorded no terminal upgrade transaction.

## Why This Change Was Made

Before managed quiescence, OCM now checks whether the current operation belongs to the target Gateway's exact Unix process group. Only that self-owned caller moves into an independent group; external callers and operations targeting another environment are unchanged. OCM fails before stopping the service if it cannot preserve the operation owner.

## User Impact

Gateway-initiated runtime upgrades can now stop their source Gateway, complete the switch, restart healthy, and leave an auditable terminal transaction instead of silently dying midway.

## Evidence

- Red on canonical base `acebb07d260792ea23684d266a87e03df4c20217`: Gateway and upgrade shared one process group; quiescence left `old-runtime`, `serviceRunning: false`, and empty upgrade history.
- Green on exact head `98ca5a6d3141594b1fd7a91ddc3d7568834c3fa6`: the upgrade owner moved to a separate process group, switched to `new-runtime`, returned with `gatewayState: running`, and persisted exactly one `switched` transaction.
- Added an end-to-end daemon regression plus matching-group, other-environment, and external-caller guards.
- Source-blind behavior validation repeated the workflow twice from compiled artifacts and passed an invalid-target negative control.
- `cargo fmt --check`, `git diff --check`, focused regressions, and the full supported Linux arm64 test surface passed. Nine existing self-update tests are intentionally unavailable on Linux arm64 and remain covered by required x86_64 CI.
- Automated review completed with no actionable findings.
- The first Ubuntu run exposed a race in the new proof harness: fast x86_64 moved the upgrade process before the test sampled its inherited group. The harness now records the inherited group before `exec`; the corrected test passes on Linux x86_64 and arm64.
- The first macOS run showed that the ad-hoc Python health fixture was not portable to that runner. The regression now uses the repository's established Node health-server pattern; no production code changed in either CI follow-up.

### DIFF
--- a/src/service/mod.rs
+++ b/src/service/mod.rs
@@ -179,6 +179,7 @@ impl<'a> ServiceService<'a> {
         if !state.running && !status.running {
             return Ok(None);
         }
+        self.preserve_operation_owner_before_managed_quiescence(name, &status)?;
         if state.running {
             self.stop_locked(name)?;
         }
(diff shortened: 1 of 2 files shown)


fuller-stack-dev:
Maintainer landing complete.

- Verified head: `98ca5a6d3141594b1fd7a91ddc3d7568834c3fa6`
- Squash merge: `e04c10166d6b58932213bf8422ccfc201cf6ac37`
- Base proof: Gateway-owned upgrade died during quiescence, leaving the old runtime stopped with no transaction.
- Head proof: the same workflow switched to the new runtime, returned healthy, and recorded exactly one terminal `switched` transaction.
- Regression coverage: full daemon lifecycle, adjacent-environment guard, external-caller guard, and invalid-target negative control.
- Required CI: Ubuntu, macOS, Windows, Rust 1.88, formatting, and CodeQL all passed.

PR #136, opened 2026-09-02
fix: bound git worktree helper subprocesses

SebTardif:
## What Problem This Solves

Fixes an issue where `ocm` dev, setup, and upgrade would hang forever when a git worktree helper (`worktree add` / `remove` / `status` / `ls-files` / `submodule foreach`) blocked on a lock or a stuck child. Those helpers called `Command::output()` with no kill deadline.

The first timeout patch waited only on the direct child, then joined stdout/stderr readers with no deadline. A descendant that still held those inherited pipes could block the caller after the child had already exited. The drain path then called terminate on that already-exited child; `try_wait()` returned `Some` and skipped process-group KILL, so a TERM-resistant descendant stayed alive.

On origin/main the hang is here:

https://github.com/openclaw/ocm/blob/e04c10166d6b58932213bf8422ccfc201cf6ac37/src/openclaw_repo.rs#L123-L129

https://github.com/openclaw/ocm/commit/e04c10166d6b58932213bf8422ccfc201cf6ac37

## Why This Change Was Made

Production worktree helpers now go through a shared timed runner. The child is polled, then terminated (process group TERM, then KILL) when the deadline expires. The same deadline stays active while the captured pipes drain. Cleanup waits for the process group, not only the direct child, so a descendant that ignores TERM still gets KILL. Fixture git used only to build temp repos stays unbounded. The 15s default is unchanged.

## User Impact

A wedged git during worktree setup or cleanup fails after 15s instead of blocking the CLI. A helper whose direct child already exited no longer hangs on leftover pipe holders, and TERM-resistant leftovers are killed with the group. Successful worktree add/remove/status behavior is unchanged.

## Evidence

Live analog of the remaining descendant leak. The grandchild ignores SIGTERM and holds a pipe. TERM alone leaves it alive. KILL removes it:

```console
$ python3 /tmp/proof-ocm136-p2-live.py
descendant_pid=37239
alive_after_term_only=True
gone_after_kill=True
DONE analog
```

The patched `command_output` runs that same shape (parent exits, TERM-resistant descendant holds stdout) with an 800ms deadline. It returns in 1.50s and the descendant is gone. A real `ensure_openclaw_worktree` add/reuse/remove on a temp OpenClaw fixture returns in 1.02s through the same `git_output` path. The 15s production default is unchanged.

```console
$ cargo test --offline --lib git_timeout_kills_term_resistant -- --nocapture
test openclaw_repo::tests::git_timeout_kills_term_resistant_descendant ... ok
finished in 1.50s

$ cargo test --offline --lib owned_worktree_uses_canonical -- --nocapture
test openclaw_repo::tests::owned_worktree_uses_canonical_repository_paths ... ok
finished in 1.02s
```

Earlier live analog of the unbounded drain hang, plus rustc `/bin/sleep 30` vs the 200ms deadline, remains in the prior evidence.

## Real behavior proof

- **Behavior or issue addressed:** Git worktree helpers no longer block `ocm` forever. After the direct child exits, inherited stdout/stderr drains stay under the same deadline. A TERM-resistant descendant that still holds those pipes is killed with the process group.
- **Real environment tested:** macOS (Darwin 25.6.0 arm64), rustc 1.98.0, perl 5, ocm checkout `/private/tmp/ocm136-p1` on `fix/git-worktree-timeout`.
- **Exact steps or command run after this patch:** Ran `python3 /tmp/proof-ocm136-p2-live.py` to show TERM-only leaves the descendant alive. Then ran the patched `command_output` against `perl` that forks, ignores SIGTERM, writes its pid, and sleeps 30s. Then ran production `ensure_openclaw_worktree` add/reuse/remove on a temp OpenClaw git repo.
- **Evidence after fix:** terminal output above. `alive_after_term_only=True`, `gone_after_kill=True`. The patched runner returned in 1.50s with the descendant gone. `ensure_openclaw_worktree` completed add/reuse/remove in 1.02s. `GIT_COMMAND_TIMEOUT` is still 15 seconds.
- **Observed result after fix:** The helper returns while a 30s descendant would still be holding stdout. Group KILL removes a SIG_IGN leftover. A normal worktree add through `git_output` still succeeds. The 15s default is unchanged.
- **What was not tested:** A live `index.lock` hang inside a full OpenClaw checkout during `ocm setup`, the Windows job-object kill path, and lengthening the 15s default for valid slow worktree operations. That last item is an owner decision.

## Summary

Shared timed runner for git worktree helpers. Deadline covers the direct child wait and the inherited pipe drain. Process-group cleanup escalates to KILL even when the direct child has already exited. 15s default unchanged.

### DIFF
--- a/src/infra/process.rs
+++ b/src/infra/process.rs
@@ -1,6 +1,8 @@
 use std::collections::BTreeMap;
 use std::path::Path;
-use std::process::{Command, Stdio};
+use std::process::{Child, Command, ExitStatus, Output, Stdio};
+use std::thread;
+use std::time::{Duration, Instant};
 
 pub fn run_direct(
     command: &str,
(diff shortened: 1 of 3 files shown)


SebTardif:
@clawsweeper re-review

P1 keep-deadline-while-draining-pipes is now in the tip.

`command_output` no longer unbounded-joins stdout/stderr after the direct child exits. The original deadline stays active across the drain. If a descendant still holds those pipes, the process group is terminated so the readers can finish.

Live analog: after the middle process exits, a raw `select` on the inherited stdout pipe is still unread at 505ms (`raw_pipe_ready_in_500ms=False`). The patched runner returns in 0.41s on the same shape.

The 15s default is unchanged. The other P1 (preserve valid worktree operations that exceed 15 seconds) is an owner default decision and is not in this commit.

SebTardif:
@clawsweeper re-review

P2 escalate-cleanup-when-descendants-survive is in the tip.

`terminate_child` no longer returns when the direct child has already exited. It waits for live process-group members, then sends KILL. A SIGTERM-ignoring descendant that still holds stdout is gone after the runner returns (1.50s on an 800ms deadline). A real `ensure_openclaw_worktree` add/reuse/remove still finishes in 1.02s.

Live analog: `alive_after_term_only=True`, `gone_after_kill=True`.

The 15s default is unchanged.

Owner decision still needed: should valid worktree operations that take longer than 15 seconds keep running (progress-based, override, or a longer default), or is a hard 15s cutoff the intended policy after upgrade? This PR does not change that default.

steipete:
Thanks for working through the pipe-drain and descendant-cleanup cases. I’m closing this version because the new unconditional 15-second deadline also interrupts valid worktree creation, filters, and removal. Main has no such cutoff, so this changes successful operations into failures and can leave an interrupted Git mutation to recover.

A compatible cancellation policy needs a separate design decision; this maintenance pass will preserve existing slow-operation behavior rather than land the hard cutoff.

PR #137, opened 2026-09-02
fix(supervisor): bound run_once child wait

SebTardif:
## What Problem This Solves

Fixes an issue where `ocm service --once` (the `__daemon run --once` path) would hang forever when a planned env child never exited. After spawn, `run_once` called `child.wait()` with no deadline.

On timeout, the first waiter sent TERM to the process group, then returned as soon as the group leader exited. A descendant that ignores SIGTERM stayed alive after `run_once` returned and released admission.

On origin/main the hang is here:

https://github.com/openclaw/ocm/blob/e04c10166d6b58932213bf8422ccfc201cf6ac37/src/supervisor/mod.rs#L809-L811

https://github.com/openclaw/ocm/commit/e04c10166d6b58932213bf8422ccfc201cf6ac37

This is separate from PR #136, which bounds git worktree helpers.

## Why This Change Was Made

`run_once` now waits through `wait_child_with_timeout`. The child is polled, then terminated when `SERVICE_ONCE_CHILD_TIMEOUT_MS` (15s) expires. Timeout cleanup now reuses the existing group-aware supervisor shutdown: wait for live process-group members, then KILL, then reap the leader. Long-running `run_until_stopped` is unchanged. The 15s once-mode default is unchanged.

## User Impact

A stuck env child in once-mode fails after 15s instead of hanging `ocm service --once`. A TERM-resistant leftover in that group is killed before admission is released. Children that exit on their own still report their exit status.

## Evidence

Live analog: a grandchild ignores SIGTERM. TERM alone leaves it alive. KILL removes it.

```console
$ python3 /tmp/proof-ocm136-p2-live.py
descendant_pid=631
alive_after_term_only=True
gone_after_kill=True
DONE analog
```

The patched `wait_child_with_timeout` runs that same shape (leader sleeps, descendant ignores TERM) with an 800ms deadline. It returns a timeout error and the descendant is gone. The 15s production default is unchanged.

```console
$ cargo test --offline --lib wait_child_with_timeout -- --nocapture
test supervisor::tests::wait_child_with_timeout_returns_status_when_child_exits ... ok
test supervisor::tests::wait_child_with_timeout_kills_sleep_after_deadline ... ok
test supervisor::tests::wait_child_with_timeout_kills_term_resistant_descendant ... ok
finished in 1.93s
```

Earlier rustc `/bin/sleep 30` vs naive `child.wait()` remains: naive still running after 1006ms, timed waiter returns at 200ms and the sleep pid is gone.

## Real behavior proof

- **Behavior or issue addressed:** `ocm service --once` no longer blocks forever on `child.wait()`. After the 15s deadline, the waiter kills the process group, including a descendant that ignored SIGTERM, then returns an error.
- **Real environment tested:** macOS (Darwin 25.6.0 arm64), rustc 1.98.0, perl 5, ocm checkout `/private/tmp/ocm137-p2` on `fix/supervisor-run-once-timeout`.
- **Exact steps or command run after this patch:** Ran `python3 /tmp/proof-ocm136-p2-live.py`. Then ran the patched `wait_child_with_timeout` against `perl` that forks a SIGTERM-ignoring child, writes its pid, and sleeps 30s, with an 800ms deadline.
- **Evidence after fix:** terminal output above. `alive_after_term_only=True`, `gone_after_kill=True`. The patched waiter returned a timeout in 1.93s with the descendant gone. `SERVICE_ONCE_CHILD_TIMEOUT_MS` is still 15000.
- **Observed result after fix:** Timeout cleanup uses the same group-exists loop as `stop_supervisor_child`. A SIG_IGN leftover is gone after the waiter returns. A child that exits on its own still reports success. The 15s default is unchanged.
- **What was not tested:** A live `__daemon run --once` against a saved plan and admission lock, a launchd/systemd install, and the Windows job-object kill path. Whether once-mode should keep a hard 15s default is an owner decision.

## Summary

Bounded `run_once` child wait. Timeout cleanup reuses the existing supervisor process-group shutdown. 15s default unchanged.

### DIFF
--- a/src/supervisor/mod.rs
+++ b/src/supervisor/mod.rs
@@ -5,7 +5,7 @@ use std::os::unix::fs::{MetadataExt, PermissionsExt};
 #[cfg(unix)]
 use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
-use std::process::{Child, Command, Output, Stdio};
+use std::process::{Child, Command, ExitStatus, Output, Stdio};
 use std::sync::{
     Arc,
     atomic::{AtomicBool, Ordering},
@@ -73,6 +73,7 @@ const SUPERVISED_CHILD_RUNTIME_ENV_KEYS: [&str; 4] =
 const SERVICE_EXECUTABLE_OVERRIDE: &str = "OCM_SERVICE_EXECUTABLE";
 pub(crate) const SERVICE_EXECUTABLE_IDENTITY: &str = "ocm-service-supervisor";
 const SERVICE_EXECUTABLE_IDENTITY_TIMEOUT_MS: u64 = 1_000;
+const SERVICE_ONCE_CHILD_TIMEOUT_MS: u64 = 15_000;
 const SERVICE_EXECUTABLE_IDENTITY_BUSY_ATTEMPTS: usize = 5;
 const SERVICE_EXECUTABLE_IDENTITY_BUSY_RETRY_MS: u64 = 20;
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
@@ -1067,9 +1068,11 @@ impl<'a> SupervisorService<'a> {
             // admission until the child exits instead of exposing an unseen child.
             let _admission = admit_supervisor_child_start(&spec, self.env, self.cwd)?;
             let mut child = spawn_supervisor_child(&spec)?;
-            let status = child.wait().map_err(|error| {
-                format!("failed waiting for env \"{}\": {error}", spec.env_name)
-            })?;
+            let status = wait_child_with_timeout(
+                &mut child,
+                Duration::from_millis(SERVICE_ONCE_CHILD_TIMEOUT_MS),
+                &format!("env \"{}\"", spec.env_name),
+            )?;
             child_results.push(child_run_result(&spec, status.code(), 0));
         }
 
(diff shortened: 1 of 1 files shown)


SebTardif:
@clawsweeper re-review

P2 wait-for-process-group-cleanup is in the tip.

`terminate_child` now calls the same group-aware shutdown as `stop_supervisor_child`. After TERM, it waits for live process-group members, then sends KILL, then reaps the leader. A SIGTERM-ignoring descendant is gone after `wait_child_with_timeout` returns.

Live analog: `alive_after_term_only=True`, `gone_after_kill=True`.

`SERVICE_ONCE_CHILD_TIMEOUT_MS` is still 15000.

Owner decision still needed: should `__daemon run --once` always kill children after 15 seconds, or should that deadline be an explicit diagnostic opt-in so existing long once-mode children keep running? This PR does not change that default.

steipete:
Thanks for improving the process-group cleanup. I’m closing this version because `__daemon run --once` runs the saved gateway command and waits for it to finish; a healthy gateway can legitimately run longer than 15 seconds. The proposed constant is new relative to main and would terminate that gateway, return an error, and skip later planned children.

An explicit diagnostic timeout would be a separate capability. This maintenance pass will preserve the existing once-mode lifetime contract.

PR #138, opened 2026-09-02
fix: daemon refresh completes when launched by a managed Gateway

fuller-stack-dev:
## What Problem This Solves

Fixes an issue where a managed Gateway asking OCM to refresh its background service would terminate the refresh command during the supervisor handoff. The command returned no final result, service-manager activation stopped after `bootout`, and the Gateway stayed down until an external process repaired it.

## Why This Change Was Made

OCM now applies its existing managed-Gateway operation-owner protection before replacing the background supervisor. A refresh launched inside a managed Gateway process group detaches before the old supervisor stops that group; external callers remain unchanged, and a group owner that cannot detach fails before any service-manager mutation. The guard is shared by launchd and systemd paths.

## User Impact

Agents and operators can run `ocm service refresh-daemon --acknowledge-gateway-restarts` from a managed Gateway and still receive the terminal result while the replacement supervisor brings configured Gateways back online.

## Evidence

- [Red terminal recording](https://asciinema.org/a/ySBRIZWlRmswLoEr) and [green terminal recording](https://asciinema.org/a/ZfEQk1jUSwpto4JY) using the same fixture and entrypoint.
- [Proof contract, manifests, checksums, and summary](https://gist.github.com/fuller-stack-dev/98d9f400b89251ace7bba2f8f7437e3f).
- Base `e04c10166d6b58932213bf8422ccfc201cf6ac37`: deterministic same-process-group reproduction stops after `bootout`, produces no refresh output, never reaches `bootstrap`, and records only the original Gateway start.
- Head `5b58bcc7ba2e08d6172cf5776e06fcdf9c81a422`: the same fixture reaches `bootstrap`, returns success JSON, and records the replacement Gateway start.
- Source-blind validation: same-group workflow passed twice on fresh fixtures; external-caller, missing-acknowledgement, and desired-state controls passed.
- Regression and adjacent-path tests: Gateway-owned daemon refresh, Gateway-owned runtime upgrade, owner-group detection, and explicit external daemon refresh all pass.
- Repository gates: `cargo fmt --check`, `cargo check --workspace --all-targets --locked`, and `cargo test --locked` pass with Rust 1.88.0.
- Structured autoreview: scoped clean, no accepted findings; overall correctness 0.98.

### DIFF
--- a/src/service/mod.rs
+++ b/src/service/mod.rs
@@ -179,7 +179,12 @@ impl<'a> ServiceService<'a> {
         if !state.running && !status.running {
             return Ok(None);
         }
-        self.preserve_operation_owner_before_managed_quiescence(name, &status)?;
+        preserve_operation_owner_before_managed_gateway_stop(
+            name,
+            status.child_pid,
+            self.env,
+            &format!("quiescing env \"{name}\""),
+        )?;
         if state.running {
             self.stop_locked(name)?;
         }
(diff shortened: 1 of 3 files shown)


fuller-stack-dev:
Landed as `6103d59176fbbd8e2d307f65f0e556a82e8985ff` after all required CI checks passed. Red/green proof: [base recording](https://asciinema.org/a/ySBRIZWlRmswLoEr), [exact-head recording](https://asciinema.org/a/ZfEQk1jUSwpto4JY), and [contract/manifests/checksums](https://gist.github.com/fuller-stack-dev/98d9f400b89251ace7bba2f8f7437e3f).

PR #140, opened 2026-09-03
Restore services after interrupted upgrades

fuller-stack-dev:
## Summary

- defer SIGINT and SIGTERM while an upgrade has rollback-backed state in flight
- route an observed interruption through the existing rollback path
- restore the source binding and original managed-service state before exiting
- cover both initially running and initially stopped services

## Root cause

OCM persisted `service_running=false` while quiescing a gateway for its safety snapshot, but retained the prior running state only inside the upgrade process. A SIGTERM during that interval terminated the process before rollback or restoration ran. The supervisor then obeyed the persisted stopped state, leaving the gateway offline.

The signal fence is limited to the rollback-backed critical section. Signals outside it retain normal termination behavior. SIGKILL remains inherently uncatchable.

## Tests

- red/base and green/head reproduction with the same disposable launchd-managed environment and exact SIGTERM timing
- `interrupted_upgrade_restores_a_running_service`
- `interrupted_upgrade_keeps_a_stopped_service_stopped`
- all project tests pass with three unrelated baseline failures excluded; all three reproduce on untouched canonical main
- `cargo fmt --check`
- `cargo check`
- touched-scope Clippy with warnings denied
- source-blind CLI behavior validation, including persistence and invalid-target probes
- autoreview: scoped clean

## Evidence

The proof comment contains public red/green videos, raw transcripts, contact sheets, behavior-validator results, and a manifest with exact base/head SHAs and SHA-256 checksums.

PR #141, opened 2026-09-03
Close the upgrade signal-to-commit race

fuller-stack-dev:
## Summary

- follow up on the final signal-to-commit race identified in #140 review
- combine critical depth and pending-interrupt state into one atomic value
- make handler acceptance and final fence closure mutually exclusive
- apply the same atomic commit handoff to explicit rollback transactions

## Root cause

#140 checked `interrupted()` and then called `commit()` separately. A signal handler could set the interruption flag between those operations, leaving no later observer before the transaction consumed its rollback state.

This change gives the signal handler and final commit handoff one atomic linearization point. If the handler records the interrupt first, fence closure fails and the existing rollback path runs. If fence closure wins, the critical section is already closed and a later signal follows normal post-transaction termination behavior.

## Tests

- debugger-controlled reproduction on unmodified `a9ad7a49`: SIGTERM delivered at `UpgradeTransaction::commit` is accepted but the command incorrectly reports `outcome: switched`
- `final_commit_and_interrupt_cannot_both_win` covers both orderings and 1,000 concurrent races
- `interrupted_upgrade_restores_a_running_service`
- `interrupted_upgrade_keeps_a_stopped_service_stopped`
- all remaining project tests pass with three unrelated local baseline failures excluded
- `cargo fmt --check`
- `cargo check`
- touched-scope Clippy with warnings denied
- autoreview: scoped clean

## Evidence

The operator-level red/green videos and source-blind service-state proof are attached to #140. This follow-up preserves those real SIGTERM paths and adds exact boundary coverage for the review finding.

### DIFF
--- a/src/cli/upgrade.rs
+++ b/src/cli/upgrade.rs
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::sync::{
     Arc, Mutex, OnceLock,
-    atomic::{AtomicBool, AtomicUsize, Ordering},
+    atomic::{AtomicUsize, Ordering},
     mpsc,
 };
 use std::thread;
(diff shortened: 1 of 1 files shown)


fuller-stack-dev:
@clawsweeper review

PR #142, opened 2026-09-04
fix: macOS permissions persist across OCM updates

fuller-stack-dev:
## What Problem This Solves

Fixes an issue where macOS users have to grant OCM file-access permissions again after each update. Published OCM binaries currently have linker-generated ad-hoc signatures, so their designated requirement is only the changing CDHash.

## Why This Change Was Made

The canonical release workflow now signs both macOS architectures with one Developer ID Application identity and the stable identifier `com.openclaw.ocm`. It enables the hardened runtime and secure timestamp, submits a temporary ZIP to Apple's notary service, checks Gatekeeper acceptance, and validates the executable again after the final tar archive is extracted.

Linux packaging and the existing tarball installer format are unchanged.

## Security and release setup

The certificate import action is pinned to an exact commit. Signing remains limited to the existing protected-main, signed-tag release workflow.

Maintainers must configure these repository secrets:

- `APPSTORE_CERTIFICATES_FILE_BASE64`
- `APPSTORE_CERTIFICATES_PASSWORD`
- `APPSTORE_API_PRIVATE_KEY`
- `APPSTORE_API_KEY_ID`
- `APPSTORE_ISSUER_ID`

They must also configure the repository variable `MACOS_TEAM_ID`. `docs/RELEASING.md` documents the certificate and App Store Connect API key requirements.

Apple cannot staple a ticket to a standalone command-line executable or `.tar.gz` archive. The workflow notarizes the signed executable in a ZIP submission, then uses `spctl` to verify Apple's ticket before packaging. The embedded Mach-O signature survives the tar create-and-extract round trip, which the packaging script now checks.

## User Impact

Future macOS releases have a stable code identity across updates, so TCC can retain OCM's file-access approval when the binary changes. Releases fail closed if signing, notarization, or post-extraction verification fails.

## Evidence

Current `v0.2.36` arm64 release before this change:

```text
Signature=adhoc
TeamIdentifier=not set
# designated => cdhash H"7af90e4a15e3785ea048ad03a09f395c5700cb95"
spctl: rejected
```

Local checks:

- `cargo test --locked --test release_asset_tests`: 17 passed
- `cargo check --workspace --all-targets --locked`: passed
- `cargo fmt --all -- --check`: passed
- `bash -n` for all changed release scripts: passed
- ShellCheck 0.11.0 for all changed release scripts: passed
- actionlint 1.7.12 for `.github/workflows/release.yml`: passed
- The new validator rejects the published ad-hoc `v0.2.36` binary and mocked unnotarized code.
- [GitHub CI run 33827450741](https://github.com/openclaw/ocm/actions/runs/33827450741): all required jobs passed on macOS, Ubuntu, and Windows.

The broad `cargo test --locked` run passed the 334 library tests and the release-related integration targets, but the unrelated macOS daemon test `gateway_owned_daemon_refresh_survives_its_source_process_group` failed again when rerun alone. This PR does not change Rust service or daemon code.

Live Developer ID signing and Apple notarization could not be exercised locally without the repository's Apple credentials. The release workflow will provide the final end-to-end proof once maintainers configure the documented secrets and variable.

## Worked on by

- @fuller-stack-dev

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,8 @@
 name: Release
 run-name: Release ${{ inputs.tag }}
 
+# macOS signing and notarization prerequisites: docs/RELEASING.md
+
 on:
   workflow_dispatch:
     inputs:
@@ -63,10 +65,13 @@ jobs:
         include:
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
+            macos: false
           - os: macos-15-intel
             target: x86_64-apple-darwin
+            macos: true
           - os: macos-14
             target: aarch64-apple-darwin
+            macos: true
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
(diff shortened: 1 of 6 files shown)


fuller-stack-dev:
Merged as 8f33d8953eaec036c677d3d07db5e04db4f842a4. Required CI was green at the pinned PR head. Before the next release, configure the Apple signing secrets and MACOS_TEAM_ID documented in docs/RELEASING.md, then validate the first notarized release.

PR #158, opened 2026-09-07
feat(npm): prepare signed binary distribution and OIDC publishing

vincentkoc:
## What Problem This Solves

OCM's Rust executable cannot currently be installed through npm. A wrapper alone would also allow native self-update to overwrite package-managed files, leave daemon bindings in temporary npx caches, or break gateway-owned refreshes by forwarding termination from a Node parent.

## Why This Change Was Made

Prepare one public package, `@openclaw/ocm`, with exact optional aliases to same-name platform versions. The launcher uses `process.execve`; installation has no lifecycle hooks, downloads, or Rust compiler dependency.

Add manual, protected-main OIDC publication of the already signed binary release. Preparation checks the signed source, exact-main CI, complete release inventory, original checksums, and GitHub asset digests. It generates deterministic tarballs from the tagged source, tests those exact files, then publishes platform payloads serially and the root channel last. Full preflight, SHA-512 retry identity, separate stable/prerelease platform tags, and immediate channel rechecks prevent stale publication from rolling tags backward.

Keep updates with the installation owner, refuse temporary npx daemon bindings, and protect shell-installer destinations before and after downloading. Preserve explicit daemon refresh and the existing Homebrew ownership guard.

Keep automatic-release verification synchronized with all nine CI jobs, including the four npm lanes, without relaxing exact identity, inventory, or success checks.

## User Impact

- Supported npm targets: macOS ARM64/x64 and Linux x64 glibc; Node `^22.15.0 || >=24.0.0`.
- Global/local npm installations update through their owning npm installation. `self update --check` explicitly reports GitHub release availability, not npm availability.
- npx can inspect and stop services but cannot create or rebind a daemon to its temporary cache.
- The package is not published by this PR. The first npm publication requires a separately approved future signed release, a legitimate package bootstrap, and npm-side trusted-publisher setup with direct `npm publish` permission.
- GitHub environment `npm` is configured for the `main` branch only. No tokens, tags, versions, releases, signing credentials, or automatic release settings were changed.

## Evidence

- `cargo check --workspace --all-targets --locked` passed.
- All 18 installer/self-update tests passed, including missing managed directories, symlinked parents, destinations changing during download, direct native invocation, aliases, and custom prefixes.
- npm-launched gateway refresh survived its caller process-group teardown.
- npx inspection/stop and daemon-binding refusal passed. An existing detached recovery-handoff test missed its log once in the broader local suite; its isolated rerun passed unchanged.
- 13 Python tests passed: actual isolated npm global install/upgrade/reinstall, local installation, npx, omitted optionals, argv/stdin/exit/signals, deterministic packaging, original checksum verification, integrity mismatch, partial retries, and zero-write stale-channel rejection.
- All 33 automatic-release tests passed. The updated nine-job unit and production CLI fixtures reproduced the old gate failure before the repair; missing, failed, skipped, renamed, and duplicate npm jobs remain rejected.
- macOS ARM64 installation of the existing v0.2.39 signed fixture preserved executable bytes and passed CLI, Developer ID, and online notarization checks. That fixture is not eligible for production npm publication; production preparation rejected it.
- Rust formatting, Ruff formatting/lint, actionlint, and `git diff --check` passed.
- Hosted CI adds Node-minimum coverage and all three native npm platforms. Live OIDC publication remains unverified until the separately approved npm account setup/publication.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,9 +74,45 @@ jobs:
         with:
           toolchain: stable
       - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: '22.15.0'
+          package-manager-cache: false
       - name: Run test suite
         run: cargo test --locked
       - name: Smoke test cargo install
         run: |
           cargo install --locked --path . --root "$RUNNER_TEMP/ocm-install"
           "$RUNNER_TEMP/ocm-install/bin/ocm" --version
+
+  npm:
+    name: npm (${{ matrix.os }}, Node ${{ matrix.node }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { os: ubuntu-latest, node: '22.15.0' }
+          - { os: ubuntu-latest, node: '24' }
+          - { os: macos-15-intel, node: '24' }
+          - { os: macos-15, node: '24' }
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+        with:
+          node-version: ${{ matrix.node }}
+          package-manager-cache: false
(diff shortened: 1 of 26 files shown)


vincentkoc:
Fixed the release CI inventory finding in b9a0856856bffacec43a9ed7feb3db616a7cfab7. Both nine-job success fixtures reproduced the old failure; all 33 automatic-release tests now pass, including rejection of missing, failed, skipped, renamed, or duplicate npm jobs. Exact identity/count/success checks are unchanged. Focused review, Rust formatting, Ruff lint, and `git diff --check` passed. Exact-head CI is running; automatic releases remain disabled.

PR #173, opened 2026-09-09
fix: resolve dev sources from the current checkout

shakkernerd:
## What Problem This Solves

Fixes an issue where starting a new dev environment could choose a neighboring or previously remembered OpenClaw repository instead of the checkout containing the current directory. Deeply nested directories could also miss their enclosing checkout and select another source.

## Why This Change Was Made

Resolve new dev sources from an explicit `--repo` or the enclosing OpenClaw checkout, canonicalize the selected path, and require `--repo` when neither is available. Existing environment bindings and interactive setup discovery retain their behavior.

## User Impact

New dev environments use a predictable source. Reusing the same explicit path through a symlink remains valid, and running from outside a checkout gives actionable guidance without selecting another repository or prompting.

## Evidence

- Remote `cargo fmt --check` and `cargo check --workspace --all-targets --locked` passed.
- All 78 affected remote command tests passed: 46 dev, 9 CLI invocation, 9 setup, and 14 environment destroy tests. Real Git fixtures cover deep directories, neighboring checkouts, legacy preferences, canonical path aliases, and existing environment reuse.
- Independent source review verified that setup's broader discovery remains unchanged and that canonical creation and resumption agree.
- Hosted Linux and macOS tests, formatting, Rust 1.88 compatibility, Windows compilation, npm installation checks, and CodeQL passed on the final head.


shakkernerd:
Fixed: new dev environments select the explicit or enclosing checkout while existing environments retain their recorded source.

PR #174, opened 2026-09-09
fix: prevent background services from competing with source watch

shakkernerd:
## What Problem This Solves

Fixes an issue where starting or restarting an OCM background service could schedule a competing Gateway while `ocm dev --watch` was active, including during rebuilds when its port was temporarily free.

## Why This Change Was Made

Serialize source-watch claims with service activation and actual daemon spawning. Background service installation, start, and restart are refused during watch, while forced takeover and restoration keep their existing behavior. Saved service plans cannot resurrect an expired source-watch binding.

## User Impact

OCM preserves the watch session's Gateway ownership while source commands continue to resolve to the watched checkout. README and help explain the explicit daemon refresh needed when an older background daemon remains running after an OCM update.

## Evidence

Verified remotely on an isolated Cargo target:

- Formatting and `cargo check --workspace --all-targets --locked` passed.
- All 13 `source_watch_tests` passed, including policy admission and saved-plan controls.
- Focused CLI service-activation and daemon saved-plan regressions passed.
- The existing legacy-watch fixture keeps its long-path and process-identity assertions while invoking its shell stand-in explicitly, avoiding a reproduced baseline `ETXTBSY` failure.

CI covers the full platform suite. Validation used synthetic environments and service fixtures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,13 @@ ocm dev luna --watch
 ocm dev existing-env --repo ~/src/openclaw --watch --force
 ```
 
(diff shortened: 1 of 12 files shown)


shakkernerd:
Fixed: Background services now respect active source-watch sessions while the daemon continues supervising other environments.

PR #175, opened 2026-09-09
fix: validate dev source tooling before startup

shakkernerd:
## What Problem This Solves

Fixes an issue where users starting `ocm dev` could reinstall valid source dependencies because their layout lacked `node_modules/.pnpm` or `.bin/tsx`. A borrowed source checkout could also pass preparation and stop an existing service before discovering missing source tools.

## Why This Change Was Made

Check the source command's declared tooling through resolution-only Node probes, including the watch entry point and build executable when required. Reuse valid flat or isolated dependencies. Existing owned dev bootstrap uses `pnpm install --frozen-lockfile` and checks the result; borrowed source takeover reports preparation errors before changing the service.

Configured modules directories inside the checkout are inspected before OpenClaw creates its `node_modules` link. The probe follows the native loader's configured-directory selection and local fallback. External trees remain rejected, and dependency repair refuses linked roots or stores. OpenClaw retains responsibility for the actual build and application dependency validation.

## User Impact

Developers can reuse working source dependencies without a layout-specific reinstall, including explicitly provisioned modules inside the checkout. Missing or broken tooling produces an actionable error, and failed borrowed-source preparation leaves the running service unchanged.

## Evidence

Remote Linux validation passed on the committed source:

- `cargo fmt --check`
- `cargo test --locked --test dev_command_tests dev_dependencies_` — 5 passed

The fixtures exercise real Node resolution with package entries that throw if executed. They cover flat dependency reuse, configured dependencies before linking, local fallback, frozen bootstrap and retry, missing or broken tools before takeover, and refusal to install through linked or external targets. The complete dev/source-watch suites and platform CI are the final regression gates.


shakkernerd:
Fixed: source development startup checks checkout-owned tooling before running the Gateway or taking over a service.

PR #177, opened 2026-09-09
feat: reuse matching dev watch sessions

shakkernerd:
<details>

</details>

## What Problem This Solves

Repeating `ocm dev demo --watch` currently fails even when the requested source already owns that environment. Concurrent starts can also prepare configuration before losing the watch lease, and deriving an active session's link from edited config can point callers at a different port.

## Why This Change Was Made

Reuse matching active watches before setup and retain the exact source request when another invocation wins startup. Only the lease owner prepares watch configuration. Capture the launch root and port with the source override, and use that endpoint for active dev status, Gateway resolution, one-shot source commands, and `env exec`.

## User Impact

Repeated matching watch commands return the existing status/link without installation, onboarding, or restart. Source, root, and explicit port conflicts fail; starting/restoring sessions report progress without guessing a link. An older active watch without endpoint metadata must be stopped from its original terminal before reuse. Existing service, takeover, and explicit runtime/launcher selection behavior is retained. README/help describe the behavior.

## Evidence

- Remote formatting and locked all-target compilation passed. No Rust toolchain, binary, or dependencies were run or installed locally.
- 55 focused native CLI, source-watch, help, and resolution tests passed. They cover edited launch ports, inherited execution environment, unchanged config/lease data, surviving watchers after controller exit, service exclusion, and legacy/transitional states.
- A controlled race runs two real CLI invocations: the losing invocation preserves config and explicit endpoint constraints while only one watcher starts.
- The full affected suites passed remotely: 196 integration cases and 7 library cases (203 total). Independent review, including the final `env exec` caller delta, is clean. Hosted platform CI and CodeQL passed on the final head.


shakkernerd:
Added: reuse of matching dev watch sessions with status and commands bound to their original launch endpoints.

PR #178, opened 2026-09-09
feat: stop source-watch sessions and recover owned processes

shakkernerd:
## What Problem This Solves

A source watch can keep running after its controlling OCM process exits, leaving cleanup and background service restoration unfinished. There is no supported command to stop that session from another terminal.

## Why This Change Was Made

Add `ocm dev stop <env>` to request a stop from the recorded controller or recover a crashed session after verifying process ownership. Dependency checks, installation, onboarding, and Gateway children participate in cancellation; stop reports success after owned processes stop and any taken-over service is restored.

New watch claims verify the running daemon's admission capability against its native process identity. An incompatible or unverified daemon must finish starting or be explicitly refreshed before a new claim. This keeps an older daemon from starting a competing Gateway through its previous ownership-lock path.

## User Impact

Stop or retry cleanup while preserving the environment, source checkout, configuration, and dependencies. Dotted environment names have separate ownership files. Legacy or unverifiable ownership is retained with an explanatory error. Existing watch reuse and stop remain available; confirmed stopped or unloaded daemons do not require refresh.

On macOS, stop distinguishes suspension from process exit and retains an unreaped process's identity through group cleanup. Windows keeps the active lease generation readable for status and named stop, uses process jobs for child cleanup, and retains a pending session when crash cleanup cannot be verified.

## Evidence

- The exact final affected remote suite passed 207 cases: 41 daemon, 60 dev, 14 destroy, 32 CLI/help, 39 service, 4 admission guards, 16 source-watch, and 1 completion-order regression.
- The daemon guard and Windows corrections passed remote formatting, Linux all-target compilation, Windows GNU all-target compilation, Node syntax, and targeted native Linux checks.
- A separate native old-to-new daemon exercise passed: a watch refused the old running daemon, the supported explicit refresh preserved desired state, and the refreshed daemon admitted a real source watcher that stopped cleanly. All fixture processes were reaped and the fixture was removed.
- Exact-head CI and CodeQL passed. Native macOS completed the full test suite and installation smoke test, including the process-identity and stop/orphan/terminal regressions.
- Independent source and integration review is clean.
- Native Windows passed both lock compatibility regressions and all four lifecycle scenarios, covering active status, named/repeated stop, crash cleanup, mismatched identity, and unverified pending ownership. The fixture verified process cleanup and removed its temporary root.


shakkernerd:
Added: `ocm dev stop <env>` stops or recovers owned source-watch sessions and restores previously running services while preserving the environment and checkout.

PR #179, opened 2026-09-09
fix: stop dev sessions before destroying environments

shakkernerd:
## What Problem This Solves

Fixes an issue where destroying an environment could remove state while its foreground source-watch session still owned processes or needed to restore a managed service.

## Why This Change Was Made

Destroy uses the recorded session generation to stop and verify its owner, then rechecks the environment binding and protection before continuing teardown. Running-watch previews explicitly defer process inspection until shutdown. Direct removal, pruning, and simulation cleanup share the stopped-session guard, and record cleanup follows the existing operation/admission/registry lock order.

Depends on #178.

## User Impact

`ocm env destroy <env> --yes` safely completes dev shutdown before removing the environment. A replacement session or changed binding preserves state. `env remove`, `env prune`, and token-guarded destruction require `ocm dev stop` first; guarded destruction then needs a fresh preview. Completed records are cleared while reusable lock files remain.

## Evidence

- Remote `cargo fmt --check` and `cargo check --workspace --all-targets --locked` passed.
- All192 tests in the dev-command, source-watch, environment-destroy, store-flow, service-command, and launcher-help suites passed on the updated integration.
- Native command cases cover ordinary destruction, forced service takeover/restoration, preview/token refusal, direct remove/prune refusal, a protection change during shutdown, and legacy sessions without usable stop ownership.
- Independent source review passed after correcting a lock-order inversion and deferring the changing watch-process snapshot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,8 @@ Use `ocm dev stop <env>` from another terminal to cancel an owned source-watch s
 
 After a watch controller crashes, `dev stop` verifies its recorded process start identity and ownership before recovering the session. It retains an unfinished session when cleanup or service restoration cannot be verified, and rejects a stale generation or a different boot/process namespace. Older watches without ownership records must be stopped from their original terminal. Windows uses the existing process job for owned child cleanup; a controller crash before child ownership is published cannot be verified automatically and requires operator recovery. Updating OCM does not add ownership records to a watch that is already running.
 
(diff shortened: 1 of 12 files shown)


shakkernerd:
Fixed: environment destruction now verifies development shutdown before removing state.

PR #182, opened 2026-09-10
fix: wait for watch ownership before Node preload hooks

shakkernerd:
## What Problem This Solves

Fixes an issue where `NODE_OPTIONS` preload hooks could run before `ocm dev --watch` finished recording the watched process's ownership.

## Why This Change Was Made

The Unix startup wait now runs before Node starts. An anonymous pipe keeps the gate usable when its descriptor number exceeds nine. The watcher retains its existing arguments, stdin, Node options, and descriptor cleanup.

## User Impact

Watched source and preload hooks begin only after OCM publishes ownership. Windows keeps its existing suspended startup and process job behavior.

## Evidence

- The new real-Node regression fails on unchanged main because a preload executes before the startup gate is released, then passes with this change.
- Both `--require` and `--import` work after release with descriptors above nine. The same fixture verifies stdin and argument preservation and prevents execution when the gate closes without release.
- Linux formatting, all-target compilation, and 12 focused tests passed, including watch preparation, terminal ownership, service restoration, and daemon admission. The high-descriptor fixture also passed all 338 library tests at default parallelism.
- [Full platform CI passed](https://github.com/openclaw/ocm/actions/runs/34442082650), including Linux and macOS tests, Windows validation, Rust 1.88 compatibility, and npm checks.


shakkernerd:
Fixed: watched Node commands now wait for recorded process ownership before running preload hooks.

PR #183, opened 2026-09-10
fix: preserve dev output after terminal or log errors

shakkernerd:
## What Problem This Solves

Fixes an issue where watched dev sessions could lose remaining output or close a child's output pipe when writing to the terminal or a log file failed.

## Why This Change Was Made

Continue draining the child's output to EOF after a destination fails, preserve the complete output in the healthy destination, and report the original write error afterward.

## User Impact

A broken terminal connection no longer truncates the dev log, and a failed log write no longer discards the remaining terminal output.

## Evidence

- Remote Crabbox formatting and locked workspace/all-target compilation pass.
- The regression covers healthy output, terminal write/flush failures, a read-only log, and simultaneous destination failures. Main stops after 8,192 of 24,577 bytes; the fix drains the full stream and preserves the healthy destination.
- The existing CLI takeover test passes and verifies both recorded dev logs.
- Independent review found no blocking issues.

### DIFF
--- a/src/cli/dev.rs
+++ b/src/cli/dev.rs
@@ -3580,18 +3580,34 @@ where
     W: Write,
 {
     let mut buffer = [0_u8; 8 * 1024];
+    let mut issue = None;
+    let mut terminal_writable = true;
+    let mut log_writable = true;
     loop {
-        let count = input.read(&mut buffer)?;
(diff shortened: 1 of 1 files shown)


shakkernerd:
Fixed: watched dev output continues to the healthy destination after a terminal or log write fails.

PR #184, opened 2026-09-10
refactor: make dev fixture modes explicit

shakkernerd:
## What Problem This Solves

Dev fixtures for source selection, preparation, cleanup and status do not consistently state whether they require a plain Gateway or a watcher. That makes mode dependencies hard to distinguish from tests of CLI defaults.

## Why This Change Was Made

Add small `dev_plain` and `dev_watch` argument builders to the existing test support and use them in fixed-mode fixtures. They preserve the original arguments and order, including the explicit `--watch` flag. Default provisioning, explicit onboarding/watch, service, flag conflicts, help and stop/status commands remain raw.

## User Impact

Test intent is explicit. CLI behavior and available flags are unchanged; existing command execution, terminal handling, assertions and cleanup are preserved.

## Evidence

- All 73 converted starts expand to their original argument token sequences.
- Remote formatting, workspace/all-target compilation and seven existing cases passed, covering plain setup, watch reuse, preparation cancellation, terminal ownership, destruction, daemon admission and status. Final formatting/static preflight also passed after narrowing the caller changes.
- Final hosted CI remains required.

### DIFF
--- a/tests/dev_command_tests.rs
+++ b/tests/dev_command_tests.rs
@@ -19,8 +19,9 @@ use ocm::supervisor::{SupervisorRuntimeChild, SupervisorRuntimeService, Supervis
 use serde_json::Value;
 
 use crate::support::{
-    TestDir, enable_fake_daemon_gateway_admission, install_fake_service_manager, ocm_env,
-    path_string, run_ocm, stderr, stdout, write_executable_script,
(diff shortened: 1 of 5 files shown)


shakkernerd:
Improved: dev fixtures now state their intended foreground mode while preserving the original command arguments.

PR #185, opened 2026-09-10
fix: block restores and upgrades during dev watches

shakkernerd:
## What Problem This Solves

Fixes an issue where snapshot restoration or an upgrade could replace environment state while a watched development session still owned it. An upgrade could also change runtime files before discovering that rollback could not safely restore the live environment.

## Why This Change Was Made

Use one source-watch admission check before snapshot restoration, ordinary upgrades, batch preparation, and explicit rollback verification. The existing operation locks prevent a new watch from starting between the check and the transaction.

## User Impact

Active, transitional, unfinished, or unreadable watches leave environment state and service policy intact. Request shutdown of recorded ownership with `ocm dev stop <env>`. Older watches without an unfinished ownership record require their original dev terminal. Unreadable or unverified ownership requires verified operator recovery, including checking the watch processes and service policy before retrying.

## Evidence

- The snapshot regression reproduces on main: restore previously succeeded during a starting watch.
- Recovery guidance covers active ownership with valid, missing, closed, and malformed session records; the full command-help suite passes.
- The upgrade refusal table verifies no OpenClaw invocation, registry/config/runtime mutation, or snapshot/history creation for single, batch, and explicit rollback commands.
- Remote Crabbox formatting and locked all-target compilation pass. Refusal/retry, direct environment/storage restore, normal rollback, failed-upgrade recovery, batch ordering, and dry-run controls pass.
- Independent source and caller review found no remaining blocker in the corrected patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,10 @@ ocm upgrade simulate mira --to beta --scenario all
 ocm upgrade simulate mira --to ./openclaw
 ```
 
+Live upgrades and rollbacks require a completed source-watch session. Run
+`ocm dev stop <env>` first; OCM refuses active or unfinished ownership before
+changing environment state, runtime files, or upgrade history.
+
 `upgrade` stages the target runtime, validates the checkpoint source, and
 prepares runtime recovery while the current managed gateway remains available.
 Preparation failures leave the source environment and service unchanged. OCM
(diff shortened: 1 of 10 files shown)


shakkernerd:
Fixed: snapshot restores, upgrades, and rollbacks now preserve environment state while development ownership remains active or unverified.

PR #186, opened 2026-09-10
fix: preserve private config access on Unix

shakkernerd:
## What Problem This Solves

Fixes an issue where updating a private OpenClaw config on Unix could replace it with a more broadly readable file. On macOS, clearing an inherited ACL after creation also left a window in which another user could open the empty replacement and retain access to bytes written later.

## Why This Change Was Made

The scoped writer stages an exclusive private replacement, verifies its protection before serialization, and uses the existing replacement path. Linux retains private file modes. macOS supplies mode `0600` and an empty ACL with inheritance disabled in the initial native creation call. Files with other authored access rules keep the existing writer.

## User Impact

Updates preserve private Unix config access while retaining the command's existing changes to config values, auth and SecretRefs.

## Evidence

- Remote formatting, Linux/Windows GNU all-target compilation, private-mode and minimum-config units, and the real dev-command rewrite passed.
- Existing config/store/clone coverage and the original private-mode failure control remain applicable.
- macOS CI runs one isolated native case with synthetic data: an unprivileged reader demonstrates the prior retained-descriptor problem, then is denied before and after the production writer writes. Compilation runs as the normal runner; only this exact test uses a privileged parent to start that reader.
- Native macOS results and current hosted checks/review remain required before landing.


shakkernerd:
Secured: private Unix config updates preserve file permissions and exclude inherited macOS access at creation.

PR #187, opened 2026-09-10
fix: retain source watch ownership after unverified completion

shakkernerd:
## What Problem This Solves

Fixes an issue where an interrupted source watch could discard child ownership before output completion was verified, allowing a later stop, reuse, service start, or destroy to treat an uncertain process tree as finished.

## Why This Change Was Made

The existing watch owner now collects both output streams before clearing its child. Read failures, incomplete output, raw termination signals, and Unix controller loss retain the generation and its error across retries. Fully observed ordinary failures remain retryable, and Windows retains its process-job cleanup behavior.

New watch generations use a distinct watch-only record and the existing OCM daemon capability check. Released legacy watch records keep their recovery rules.

## User Impact

A watch whose cleanup cannot be verified preserves its environment and ownership instead of restoring a service or allowing destructive reuse. Watched output remains visible and is recorded in the environment logs. Interactive setup keeps its real terminal; successful setup remains supported, while cancellation or failure without observed output stays unfinished.

## Evidence

- The final remote affected run passed formatting, all-target compilation, and 598 tests across the library and nine affected integration suites.
- Native worker fixtures cover held output, a signaled intermediary with private pipes, and controller loss. The worker stays alive through refused stop/service/reuse/destroy attempts until fixture-owned cleanup.
- Real-terminal controls distinguish successful setup from cancellation acknowledged with exit zero. Released legacy recovery and forced-shutdown controls remain covered.
- Hosted Linux and macOS suites, Rust 1.88, and CodeQL passed. Native Windows checks passed both byte-lock controls and the four-scenario stop lifecycle fixture.


shakkernerd:
Fixed: source watches retain unfinished ownership when process or output cleanup cannot be verified, preventing unsafe reuse, service restoration, or environment removal.

PR #188, opened 2026-09-10
fix: preserve dev bindings during snapshot restore

shakkernerd:
## What Problem This Solves

Fixes an issue where users restoring a snapshot of a dev environment would lose its source binding or revert the runtime or launcher they selected after the snapshot.

## Why This Change Was Made

Manual restore retains the dev environment's current source binding and runtime/launcher selection while restoring saved state. Upgrade rollback retains the dev identity while restoring the recorded runtime/launcher selection. Both checkpoint and legacy archive restores use the same policy.

## User Impact

Restoring dev state preserves the selected source checkout and execution binding. Ordinary runtime and launcher environments keep their existing snapshot restore behavior.

## Evidence

CLI cases cover source-only, runtime and launcher dev bindings; ordinary and legacy snapshot restore; active-watch refusal; failed restore acceptance; runtime rollback and reversal; launcher rollback; and automatic failed-upgrade recovery. Formatting, all-target compilation and the affected snapshot/upgrade integration checks passed on current main. Hosted platform checks remain the final validation step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,9 @@ safety snapshot and a linked history transaction before it stops a managed
 service or replaces runtime bytes. If restore or verification fails, OCM puts
 the pre-rollback runtime and environment state back. Rolling back the linked
 transaction safely reverses the rollback.
+Dev environments retain their current source binding during restore.
+Manual snapshot restore also keeps their current runtime/launcher binding;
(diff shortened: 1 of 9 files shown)


shakkernerd:
Fixed: restoring dev snapshots preserves source bindings while keeping the correct execution selection for manual restore and upgrade rollback.

PR #189, opened 2026-09-10
fix: preserve private access on Windows config updates

shakkernerd:
## What Problem This Solves

Fixes an issue where updating a private OpenClaw config on Windows could give the replacement file broader access inherited from its environment directory.

## Why This Change Was Made

Private replacements are created with a protected ACL and the current user as owner. OCM verifies the applied access before writing config data, then uses the existing atomic replacement path. Files with other authored access rules keep the existing writer behavior.

This builds on the shared config writer in #186.

## User Impact

Existing user-only config access survives OCM config updates. Auth settings and SecretRefs continue to follow the requested command's existing update rules.

## Evidence

- The existing native Windows regression covers a directory with inheritable access, rejection of an unprotected empty staging file, private replacement, preserved auth settings, and the ordinary writer fallback.
- That regression passed with the unchanged Windows helper and test on `9f0092b2`; the Windows CI job runs it again for this PR.
- Formatting, Linux and Windows GNU all-target checks, and three focused config/CLI cases passed on the extracted source. Fresh native Windows CI and review run on the published head.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,8 @@ jobs:
       - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
       - name: Check Windows targets
         run: cargo check --workspace --all-targets --locked
+      - name: Prove private config replacement
+        run: cargo test --locked --lib private_config_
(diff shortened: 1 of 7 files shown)


shakkernerd:
Fixed: OCM config updates preserve protected current-user access on Windows.

PR #190, opened 2026-09-10
fix: verify watched dependency script completion

shakkernerd:
## What Problem This Solves

A watched dependency install can return an ordinary error, or succeed after an optional build fails, even when a lifecycle script was interrupted. Treating only the installer exit and outer output completion as proof can release watch ownership too early.

## Why This Change Was Made

Watch-owned pnpm installs now request the existing lifecycle reporter and track complete script start/output/exit records. Interrupted, unreadable, or unfinished observations retain ownership; fully acknowledged ordinary failures remain retryable. The observer bounds individual records and active lifecycle state while continuing to stream output without a lifetime output cap.

The existing cleanup owner collects both streams before the observer result can allow child ownership to clear. Verified native Windows Job cleanup remains authoritative after cancellation.

## User Impact

Build output and diagnostics remain readable, including ordinary multiline JSON, while reporter metadata stays private. Installers retain interactive stdin when run from a terminal; onboarding keeps its existing terminal behavior. The normal watch stop, retry, service, and destruction guards continue to handle uncertain cleanup.

## Evidence

- The current parser correction passed remote formatting, all-target compilation, the reporter tests and finite CLI regression. Ordinary JSON exceeding 4 MiB, continued objects/arrays, adjacent ordinary JSON and reports, partial EOF output, and reporter prefixes split at the buffer boundary are covered. Genuine malformed, unfinished, or oversized pnpm reports remain unsafe.
- The affected remote suite passed 614 cases (357 library and 257 integration) with formatting/all-target checks and the same qualified CLI binary before and after the run.
- Real pnpm 12.3.4 native and package JavaScript entries passed ten success, ordinary-failure/retry, lifecycle-signal, optional-signal, and installer-signal scenarios. Four baseline controls reproduced premature completion while a worker remained alive. The final parser additionally passed both real entry paths for success and existing-worktree failure/retry, preserving large ordinary output and JSON adjacent to reports. A controlled oversized-output counterexample reproduced the earlier parser failure. Every fixture was cleaned up.
- A real OpenClaw at `61f761464` frozen install produced 6,735 reports and 11 matched lifecycle starts/exits, with no invalid, unmatched, signalled, or unfinished observations. The largest report was 122,801 bytes; peak active state was 3 lifecycles / 779 identity bytes, below the 4 MiB and 4,096-lifecycle bounds. Cleanup left no fixture processes.
- The final head passed hosted platform checks, including macOS/Linux tests and Windows compilation/native stop coverage, and ClawSweeper rated it Platinum with no actionable findings. The final main integration also passed formatting, all-target compilation and the lifecycle CLI regression.


shakkernerd:
Fixed: watched dependency installs retain ownership until lifecycle script completion can be verified.

PR #191, opened 2026-09-10
feat: stop and reuse plain foreground dev sessions

shakkernerd:
Builds on #190.

## What Problem This Solves

Plain `ocm dev` runs execute setup and the Gateway outside the foreground session lifecycle, so named stop cannot cancel them and overlapping invocations can repeat setup.

## Why This Change Was Made

Acquire the existing foreground owner before setup and run the native one-shot Gateway through the same process and completion handling as watch mode. Persist the selected mode with that generation, retain released session compatibility, and reject incompatible daemons before creating a new environment.

## User Impact

Users can stop plain or watched dev sessions with `ocm dev stop <env>`, reuse a matching active session without rerunning setup, and distinguish active ownership from backend watching in status. Changing mode requires stopping the existing session; named stop preserves the environment and source checkout. Plain startup now uses the same daemon compatibility guard as watch mode. An older running daemon requires the documented, explicitly acknowledged refresh, which restarts its managed Gateways.

## Evidence

Final integration is `f17aa68` on main at `0c30947`.

- Formatting, locked workspace/all-target compilation, a fresh CLI build and four focused cases passed on this exact source tree. They cover plain reuse/named stop, replaced-source refusal, and both destruction controls.
- With OCM `e7393f5` and real [OpenClaw source at `61f7614`](https://github.com/openclaw/openclaw/commit/61f76146412c0ce01e88bbbfcd25819d383c3618), fresh plain startup, matching reuse, named stop and repeated stop passed. The current client also stopped a live watch-v2 session created by the older c48 build. Source comparison preserves these foreground paths in the final integration.
- That native run also exercised read-only refusal under an admission-version-7 daemon, acknowledged refresh to version 8, and plain startup/reuse/stop afterward, preserving legacy session and desired-state bytes. It used an empty managed-Gateway plan; running-Gateway restart and interruption acknowledgement have separate CI coverage.
- Environment/config/source preservation and process cleanup passed. Disposable environments and worktrees were removed after clearing their enumerated generated build outputs; original source inventory was restored with no fallback signals.

[CI](https://github.com/openclaw/ocm/actions/runs/34493368859) passed Linux, macOS, Windows, minimum Rust, formatting and npm checks; [CodeQL](https://github.com/openclaw/ocm/actions/runs/34493364988) also passed. The affected integration suites ran without filters, and native Windows completed all five lifecycle cases.


shakkernerd:
Added: Plain foreground dev sessions support named stopping and reuse of a matching active session.

PR #192, opened 2026-09-10
fix: keep new environment state out of dev worktrees

shakkernerd:
## What Problem This Solves

Fixes an issue where users creating, cloning, or importing an environment could write state into an existing registered dev worktree by placing the destination inside it. Failed clone or import cleanup could also remove a destination symlink authored inside the checkout even when that link pointed elsewhere.

## Why This Change Was Made

Validate destination placement against existing registered dev sources before the store writes or copies state. Resolve path aliases and compare filesystem identity; also check a final destination symlink's parent because cleanup removes the entry itself. Source inspection failures stop the operation, with documented guidance for checking and recovering the source path.

The scope is new environment destinations and existing source directories. Missing-source reservation and snapshot restore retain their existing behavior.

## User Impact

Overlapping destinations fail with guidance to choose a separate environment root, preserving checkout files and the environment registry. Separate roots, including roots elsewhere in the originating repository, continue to work. Existing Git identity checks protect replacement environment state when an older dev owner is removed.

## Evidence

- Current-main command reproduction confirmed that `ocm env create overlap --root <registered-worktree>/nested-state` succeeded and wrote state into the checkout.
- The same command now refuses the overlapping root with the source and environment registry unchanged and no state written into the checkout.
- Crabbox: `cargo fmt --check`, `cargo check --workspace --all-targets --locked`, the create/clone/import command matrix, and the existing missing-worktree test extended with replacement-state removal refusal all passed.
- All 180 affected CLI, dev, clone, import, migration, start, and store tests passed in Crabbox. Independent source and platform reviews found no blockers.
- [Hosted checks](https://github.com/openclaw/ocm/pull/192/checks) cover Linux and macOS tests, Windows compilation, Rust 1.88 compilation, and formatting.


shakkernerd:
Fixed: environment creation, cloning, and import now preserve existing registered dev worktrees by refusing overlapping destinations.

PR #193, opened 2026-09-10
fix: create owned worktrees with relative Git paths

shakkernerd:
## What Problem This Solves

Git for Windows rejects the canonical verbatim absolute destination passed when OCM creates an owned development or simulation worktree. Creation fails before the requested command starts.

## Why This Change Was Made

Pass the owned destination relative to the existing `git -C` repository. Keep canonical stored paths and the existing registration, checkout-identity, reuse, and cleanup checks.

## User Impact

Owned worktree creation supports canonical Windows repository paths without changing source ownership or removing authored files.

## Evidence

- Crabbox formatting and locked all-target compilation passed.
- A real Git test covers a canonical repository path with spaces, creation, reuse, removal, and preserved source bytes; four existing cleanup controls also pass.
- Native Windows CI passed the canonical-path create/reuse/remove regression, the existing native dev-stop case, and Windows lock and private-file controls.
- Linux, macOS, MSRV, formatting and CodeQL checks passed on the prior equivalent patch. Current-main CI and exact-head review are required before landing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,8 @@ jobs:
         run: cargo check --workspace --all-targets --locked
       - name: Prove private config replacement
         run: cargo test --locked --lib private_config_
+      - name: Prove canonical owned worktree paths
+        run: cargo test --locked --lib openclaw_repo::tests::owned_worktree_uses_canonical_repository_paths -- --exact --nocapture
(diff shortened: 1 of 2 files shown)


shakkernerd:
Fixed: OCM-owned development and simulation worktrees now support canonical Windows repository paths while preserving safe reuse and cleanup.

PR #194, opened 2026-09-10
feat: initialize persistent auth for new dev environments

shakkernerd:
## What Problem This Solves

New dev environments previously started without persistent Gateway authentication in their initial config. A Gateway restart could then invalidate paired clients' authentication.

## Why This Change Was Made

OCM creates an OS-random token in a private initial config before registering a new dev environment. Publication never replaces an existing config, including one created concurrently. Minimum setup also skips writes when the resulting config is unchanged.

## User Impact

Fresh dev environments keep stable Gateway authentication across restarts and source rebuilds. Existing configs, authored auth modes, and SecretRefs retain their existing behavior; unchanged JSON5 config bytes survive repeated setup.

## Evidence

- Focused coverage checks token format and uniqueness, private initial access, stable auth after updates, existing-config preservation, and no-clobber publication.
- Store and actual dev-command coverage check config availability, ordinary environment behavior, and token absence from metadata and command output.
- Formatting, Linux and Windows GNU all-target checks, and all seven focused cases passed on the integrated source. Native Windows/macOS behavior and full CI run on the published head.


shakkernerd:
Merged as ee8deca07292de5f650a47ac3eb2316d79a8d75b. New dev environments now receive persistent Gateway token auth in a private initial config, while existing authentication is preserved. Linux, macOS and Windows checks, CodeQL, and exact-head review passed.

PR #195, opened 2026-09-10
fix: preserve Gateway operations during startup

shakkernerd:
## What Problem This Solves

Fixes an issue where an OCM daemon refresh, snapshot, or upgrade started inside a managed Gateway could be terminated by the supervisor’s Gateway process-group shutdown when that Gateway had started running but its PID was not yet visible to OCM.

## Why This Change Was Made

Both preservation callers now use the existing Gateway admission lock to read the published PID and preserve the calling process before shutdown. The lock is released before stop, restart, or convergence work. Existing process-group checks and the Windows no-op remain unchanged.

## User Impact

Gateway-started maintenance commands move out of the Gateway process group before the supervisor stops it, including during this startup interval. Snapshot and upgrade flows retain their existing service restoration policy.

## Evidence

- Two Linux regressions invoke the real OCM commands while admission is held and no Gateway PID is published, then verify preservation through group termination and snapshot service restoration.
- Both unchanged controls fail at the intended boundary against the earlier implementation: refresh reaches shutdown before publication; snapshot retains its caller in the Gateway group at shutdown.
- Formatting, all-target compilation, and the focused command controls passed after a fresh OCM build. Existing daemon refresh (native and npm), upgrade, and snapshot controls passed on Linux and macOS.
- [CI](https://github.com/openclaw/ocm/actions/runs/34473095024) passed, including Rust 1.88, Windows checks, full Linux/macOS suites, and npm installation checks.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -344,6 +344,12 @@ ocm service status mira
 
 Use `service install` when you want a background service for an environment that was created with `--no-service`, or when you want to bring an older env under background management later.
 
+On Unix, OCM waits for Gateway PID publication before moving a matching caller
+out of the Gateway process group during daemon refresh, snapshot, or upgrade
+maintenance. This protects the caller from the supervisor's signals to that
+group. A systemd unit stop or restart can still terminate the caller through its
+service cgroup.
+
 ### Read logs
 
 ```bash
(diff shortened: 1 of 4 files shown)


shakkernerd:
Fixed: Gateway-started maintenance now waits for PID publication before leaving the process group that the supervisor stops.

PR #196, opened 2026-09-10
test: allocate complete port families for daemon fixtures

shakkernerd:
## What Problem This Solves

The Gateway-owned upgrade and refresh fixtures probed one free TCP port, then passed it as the explicit Gateway port. Startup checks the whole OpenClaw port family, so an occupied neighboring port could prevent the Gateway from starting before the lifecycle assertion ran.

## Why This Change Was Made

Let the existing `env create` allocator select an available port family. Both fake Gateways already read the resulting `gateway run --port` argument. Keep the existing upgrade, direct refresh and npm refresh assertions.

## User Impact

Daemon lifecycle tests use the supported port-selection path. Product behavior is unchanged.

## Evidence

- A controlled baseline run held the neighboring port while the base remained free: the daemon refused startup until the neighbor was released, then started without a configuration change.
- Crabbox formatting, locked all-target compilation and all three existing Gateway-owned lifecycle cases passed with an explicitly rebuilt OCM executable.
- The equivalent patch is integrated onto current main; full platform CI is required before landing.

### DIFF
--- a/tests/daemon_runtime_tests.rs
+++ b/tests/daemon_runtime_tests.rs
@@ -3,7 +3,6 @@ mod support;
 use std::collections::BTreeMap;
 use std::fs;
 use std::io::Write;
-use std::net::{Ipv4Addr, TcpListener};
 use std::path::Path;
 use std::process::{Child, Command, Stdio};
 use std::sync::{Mutex, MutexGuard};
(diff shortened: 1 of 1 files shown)


shakkernerd:
Improved: Gateway-owned upgrade and refresh tests now select an available OpenClaw port family before startup.

PR #197, opened 2026-09-10
fix: refuse stale dev service plans

shakkernerd:
## What Problem This Solves

Fixes an issue where a saved dev service plan could run an old checkout after the environment's binding or recorded worktree changed.

## Why This Change Was Made

Revalidate saved dev plans in the supervisor's existing admission step, before capability probes and process launches. The plan must still identify the current dev binding and its recorded worktree.

## User Impact

OCM refuses stale plans with instructions to restart the service and refresh its plan. Matching owned-worktree plans continue to run normally.

## Evidence

- A fresh build of main at `f9f2acc` ran the matching saved plan, then failed the new regression with `saved-source launched the stale command` against an existing checkout.
- A fresh build of this change passed all five admission tests. The regression completed the matching control and all four refusals: stale source, incorrect saved binding name, runtime rebinding, and launcher rebinding. Refused plans launched no command and retained their saved state.
- Remote Linux validation passed `cargo fmt --check`, `cargo check --workspace --all-targets --locked`, and `cargo test --locked --test source_watch_daemon_admission_tests -- --nocapture` with source and CLI hashes unchanged throughout.
- Hosted [platform CI](https://github.com/openclaw/ocm/actions/runs/34478227065) passed on Linux, macOS, and Windows.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -366,6 +366,9 @@ ocm service stop mira
 ocm service restart mira
 ```
 
+Saved dev service plans must still match the env's current binding and recorded
+worktree. If OCM refuses a stale plan, use `ocm service restart <env>` to refresh it.
+
 Normal restart is gateway-aware when `ocm service status mira` reports restart
 handoff `protocol v1`: OpenClaw records eligible active sessions and subagents,
 hands the fresh-process restart back to OCM immediately, and resumes recoverable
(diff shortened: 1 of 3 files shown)


shakkernerd:
Fixed: stale dev service plans are refused before the supervisor probes or launches their commands.

PR #198, opened 2026-09-10
fix: refuse replaced dev worktrees before execution

shakkernerd:
## What Problem This Solves

Fixes an issue where environment commands and saved service plans could run a replaced dev checkout even though resuming the environment with `ocm dev` refused it.

## Why This Change Was Made

Use the existing recorded-worktree validation when selecting dev source for commands, Gateway resolution, active source routing, and startup from saved service plans. Keep metadata mismatch checks first, and resolve onboarding commands before preparing configuration.

## User Impact

Missing or unrelated recorded worktrees are refused before source starts. Valid worktrees, explicit runtime and launcher overrides, status inspection, and restored dev bindings remain usable.

## Evidence

- Qualified baseline fixtures reproduced both the normal command-resolution bypass and a cold saved-plan launch at the same replaced path after fresh resolution refused it.
- Formatting, clean CLI builds, locked all-target compilation, and 28 focused tests passed, reusing 26 unchanged cases. The final two selectors cover cold-plan refusal with a valid cached-start control and successful restore-to-resolution with a real Owned Git worktree.
- Platform CI is tracked on this PR and remains a landing requirement.


shakkernerd:
Fixed: environment commands and saved service starts now refuse missing or replaced recorded dev worktrees.

PR #199, opened 2026-09-10
fix: keep gateways running during dev service setup

shakkernerd:
## What Problem This Solves

Fixes an issue where `ocm dev stop <env>` could not cancel preparation started by `ocm dev <env> --service`. Preparation must also preserve a managed Gateway already running while other environments update the supervisor plan.

## Why This Change Was Made

Service preparation uses the existing recorded child completion and cancellation path. The supervisor retains the saved and live launch specifications and defers restart requests during preparation. Verified completion, binding and service-policy revalidation, and managed service start share the existing operation lock. The handoff preserves newer explicit stop requests, including an idempotent stop, and avoids the temporary stop caused by a separate service installation.

## User Impact

`ocm dev stop <env>` can cancel service preparation while retaining the environment and source. OCM keeps an already-running managed Gateway and its launch plan during preparation; explicit service stop or uninstall remains effective. Supported watch and foreground records retain their recovery rules. A running daemon must support the updated preparation contract before admission.

## Evidence

Remote formatting and locked workspace/all-target compilation passed with freshly rebuilt, hash-verified executables. Targeted Rust checks cover preparation cancellation, live specification retention, concurrent plan publication, the closed-record/held-lease handoff, deferred restarts, newer service policy, supported session records, stale source rejection, and cleanup/restore integration.

Six native scenarios passed with actual OpenClaw source, real Linux daemon and Gateway processes, and an isolated service-manager adapter. They cover an older running daemon and supported watch session, current-client reuse and stop, refusal before mutation, acknowledged daemon replacement preserving the saved plan and configuration, preparation after refresh, and fresh service startup followed by repeated setup with unchanged Gateway process identities. All fixture processes and worktrees were cleaned up without forced termination or fallback removal.

Independent implementation and integration review is clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ Use `ocm dev` when you want an isolated source-run checkout with its own env roo
 
 On Unix, foreground Node commands wait until OCM records their process ownership before executing source or `NODE_OPTIONS` preload hooks. The gate preserves the command arguments, terminal input, and configured Node options after release.
 
-Before starting a new foreground session, OCM checks that a running background daemon uses
+Before starting a new foreground session or service preparation, OCM checks that a running background daemon uses
 compatible ownership locks and that its runtime record identifies the current
 process. A daemon whose compatibility cannot be verified must finish starting
 or be refreshed before the session can claim ownership. During a maintenance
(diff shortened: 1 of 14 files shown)


shakkernerd:
Fixed: dev service preparation is cancellable while preserving the running Gateway and newer service stop requests.

PR #200, opened 2026-09-10
fix: preserve registered dev sources during environment cleanup

shakkernerd:
<details>

**MUST:** Keep **Allow edits from maintainers** enabled for this PR so maintainers can help update the branch when needed.

</details>

## What Problem This Solves

Fixes an issue where users removing, pruning, or destroying an environment could delete another environment's registered dev source and Git metadata when those paths were inside the cleanup target. Destroy could also terminate workers using that source. Simulation cleanup could erase a registered source inside generated build output.

## Why This Change Was Made

Inspect registered source entry points, their required Git metadata, and the symlink paths connecting them before cleanup. Keep the environment registry locked through broad process teardown and final deletion. Apply the same repository and index selection to Git inspection and mutation.

Preserve directional removal of independent owned child worktrees, recoverable metadata for missing worktrees, and registered replacement state created at a formerly missing source path. Replacement ownership comes from the registry and applies to plain and dev-bound environments. The stored dev metadata format remains unchanged.

## User Impact

Cleanup explains which dependent dev environment must be removed first. Registered sources and their workers survive a refused cleanup, including sources reached through aliases or external Git metadata. README, usage guidance, and command help describe the boundary.

## Evidence

- Baseline `f9f2acc`: a real CLI destroy reported success after terminating a nested registered source's worker and deleting that source. A paired metadata case also reproduced deletion of required Git common metadata.
- Remote formatting, `cargo check --workspace --all-targets --locked`, and 19 focused tests passed on the corrected source. Coverage includes nine Git layouts with and without inherited selectors, plain and dev-bound replacement state through aliases, dirty and missing worktrees, partial teardown, watch stopping, and simulation cleanup.
- The replacement control reproduces unrelated-cleanup refusal on the earlier executable and passes on the corrected executable while preserving replacement bytes, registration, and detached Git history. Earlier permitted-child, Git-slot, and simulation-selector controls remain applicable.
- Two native CLI orderings passed: registration first preserves the source and worker; cleanup first holds the registry lock and rejects subsequent publication of the removed source. Both finish with an empty environment registry. This proof is reused across the replacement-only correction; lock and callback behavior is unchanged.
- Test fixtures and their process groups were removed. Full platform CI and exact-head review passed on the prepared commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,8 @@ Environment creation, cloning, and import reject roots that overlap an already
 registered dev worktree that exists on disk, including source and destination
 path aliases. Choose a separate environment root to preserve the checkout.
 
+Removal, pruning, destroy, and simulation cleanup preserve other registered dev sources and the Git metadata they need. Remove dependent dev environments before deleting their containing environment or worktree.
+
 ### Try beta or pin a specific release
 
 ```bash
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -643,6 +643,10 @@ by an older OCM without usable stop ownership must be stopped from its original
 terminal. Environment removal clears completed watch records and retains the
 reusable synchronization lock files.
 
+Cleanup refuses to remove another registered dev source or its required Git
+metadata, including sources reached through path aliases. Destroy checks before
+signalling source workers; simulation checks before discarding generated files.
+
 `destroy` is the stronger cleanup path.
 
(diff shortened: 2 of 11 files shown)


shakkernerd:
Fixed: environment cleanup preserves other registered dev sources, their required Git metadata, and source workers.

PR #201, opened 2026-09-10
fix: lock source owners during dev registration

shakkernerd:
## What Problem This Solves

Fixes an issue where creating a dev environment from source belonging to another environment could prepare a worktree and publish its binding during an operation on the source owner. Local upgrade simulations and direct store updates could register the same overlapping dependency.

## Why This Change Was Made

Builds on #200.

Registration acquires the existing operation locks for environments that can affect the selected source, including owned worktrees and their Git registration metadata. It checks identity and current owners again under the registry lock before publication. Unchanged source tuples remain writable during restore and recovery.

## User Impact

Retry new or changed dev bindings after the source owner's operation finishes. Independent sibling worktrees remain usable. Failed registration preserves reused or already-published source, and safely cleans up worktrees created by the rejected attempt.

## Evidence

The baseline CLI reproduction confirmed that a child could be published while its containing environment's operation lock was held. The regression checks rejection before source preparation, successful retry, nested owned sources, and independent sibling creation.

Remote formatting, locked workspace/all-target compilation, and 20 distinct focused cases passed across the registration and integration gates. The final source passed seven affected dev cases with a freshly built CLI, including dependency failure/retry and foreground reuse/stop. Independent implementation and integration reviews found no blockers.

[Final-head CI](https://github.com/openclaw/ocm/actions/runs/34503102157) and [CodeQL](https://github.com/openclaw/ocm/actions/runs/34503100604) passed. [ClawSweeper](https://github.com/openclaw/ocm/pull/201#issuecomment-5619684463) rated the exact head Platinum with no findings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,11 @@ Plain and watched foreground sessions support `ocm dev stop <env>` and reuse a
 matching active session. `dev status` reports backend watching separately from
 session ownership.
 
+Creating or changing a dev binding is refused while an environment containing or
+owning its source, worktree path, or required Git metadata is busy with another operation.
+This also applies to local upgrade simulations; retry after that operation finishes.
+Unchanged source bindings remain writable during restore and recovery.
+
 New dev environments receive a private Gateway token in their initial config before
 the environment is registered. The token remains stable through restarts and
 source rebuilds, so paired clients can reconnect. Initialization never replaces an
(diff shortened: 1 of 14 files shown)


shakkernerd:
Fixed: OCM now refuses new or changed dev bindings during operations on their source environments while preserving unchanged recovery writes.

PR #203, opened 2026-09-10
fix: preserve registered dev sources during restore

shakkernerd:
## What Problem This Solves

Fixes an issue where restoring or rolling back a containing environment could erase another named environment's dev checkout or Git registration while leaving its environment registration intact.

## Why This Change Was Made

Restore checks the selected checkpoint's recorded scope against other environments' registered source and Git dependencies before service quiescence and again before staging replacement. Required upgrade rollback also checks its recovery scope; explicit rollback checks both the selected checkpoint and its failure-recovery checkpoint. The check uses the operation lock already shared with source registration through restore acceptance.

## User Impact

Unsafe restores are refused before replacement. A checkpoint that preserves the complete dev source and Git metadata can still restore surrounding state. Legacy restore updates and runtime-residue cleanup preserve captured directory links without following their destinations. Full snapshots retain their behavior for the selected environment's own source and unregistered independent content, and disabled automatic rollback keeps its existing behavior.

## Evidence

- Existing-behavior controls reproduced source/Git loss during full restore and source modification through captured directory links, using isolated fixtures and real Git repositories.
- Remote Linux validation passed 50 positive cases: 10 checkpoint-scope cases, 39 snapshot cases, and the registration operation-lock case. All fixture processes and temporary state were cleaned up.
- Remote formatting and locked Linux/Windows GNU all-target compilation passed. Independent source review and the focused test correction review are complete.
- Replayed the unchanged restore commit onto the landed prerequisites with an identical range-diff. [Final hosted CI](https://github.com/openclaw/ocm/actions/runs/34505817866) passed every job, including Linux/macOS tests and Windows compilation; [CodeQL](https://github.com/openclaw/ocm/actions/runs/34505812934) also passed.
- The [current-head review attempt](https://github.com/openclaw/clawsweeper/actions/runs/34506549818) failed during the review service's manual admission step before producing a verdict. Independent review is clean; the standing reviewer-error policy applies without a claimed current rating.

### DIFF
--- a/docs/USAGE.md
+++ b/docs/USAGE.md
@@ -508,7 +508,18 @@ not add crash recovery for an uncatchable process or machine failure.
 
 This policy belongs to the environment registration. Clone and import start with
 an empty list. A separately requested full snapshot, export, or clone still
-includes independent content. Full snapshot restore still rewinds that content.
+includes independent content. Full snapshot restore still rewinds unregistered
+independent content and the selected environment's own source.
+
+Restores and required rollbacks refuse checkpoints whose recorded scope would
+replace another named environment's registered dev source or required Git
+metadata, before service quiescence or restore staging. Explicit rollback checks
+both the selected checkpoint and its failure-recovery checkpoint. `--no-rollback`
+keeps its existing behavior. The check reads only known registered source paths
+and Git identity metadata, including surviving history for a missing worktree.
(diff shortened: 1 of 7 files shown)


shakkernerd:
Fixed: restores and required rollbacks preserve other registered dev sources and captured directory links.

PR #204, opened 2026-09-10
fix: prevent intermittent snapshot lifecycle test failures

shakkernerd:
## What Problem This Solves

Resolves a problem where snapshot lifecycle checks intermittently fail with `EOF while parsing a value at line 1 column 0` while a simulated supervisor changes state.

## Why This Change Was Made

The fixture rewrote live supervisor runtime JSON with `fs::write`, allowing the snapshot command to read the file after truncation and before the new bytes arrived. All valid runtime publications in this fixture now use the existing atomic JSON helper. The starting-service observer is also stopped and joined before the snapshot result is asserted.

## User Impact

Snapshot regression checks can reliably exercise service quiescence and restoration. Shipped command behavior is unchanged, and existing lifecycle assertions and intentionally corrupt-state fixtures are preserved.

## Evidence

- On main `e181de20`, a controlled publication window reproduced the matching EOF through the real snapshot CLI. Holding an unpublished temporary file instead preserved snapshot success. Both controls passed, and the unmodified starting-service test also passed normally.
- Remote `cargo fmt --check`, `cargo check --locked --test env_snapshot_tests`, and the exact starting-service test passed on the repaired source with a freshly built CLI.
- Remote `cargo test --locked --test env_snapshot_tests`: all 39 tests passed.
- [CI for the current head](https://github.com/openclaw/ocm/actions/runs/34497873491) covers Linux/macOS tests and install smoke checks, Rust 1.88, Windows compilation, and npm checks.

### DIFF
--- a/tests/env_snapshot_tests.rs
+++ b/tests/env_snapshot_tests.rs
@@ -26,7 +26,7 @@ use serde_json::Value;
 
 use crate::support::{
     TestDir, TestHttpServer, install_fake_launchctl, ocm_env, path_string, run_ocm, stderr, stdout,
-    write_executable_script, write_text,
+    write_executable_script, write_json_replacing_path, write_text,
 };
 
 fn write_running_snapshot_service(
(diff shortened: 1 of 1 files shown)


shakkernerd:
Fixed: snapshot lifecycle tests no longer read partially published supervisor state.

PR #205, opened 2026-09-10
fix: retain dev UI addresses across restarts

shakkernerd:
## What Problem This Solves

Fixes an issue where restarting `ocm dev <env> --ui` could select a different UI address and leave the existing browser URL unavailable.

## Why This Change Was Made

Retain the UI port for the environment's name, root, and creation identity, and reserve it alongside Gateway and active UI allocations. An occupied retained address fails startup until it is free.

The reservation survives older registry writers, snapshot restore, and runtime-state cleanup. Clone and import select their own addresses; environment removal releases its reservation.

## User Impact

Stop and restart the same environment at the same UI address. Another environment cannot inherit a copied reservation.

## Evidence

- Remote formatting, all-target compilation, and a fresh CLI build passed. The earlier scoped head passed hosted Linux/macOS suites, Rust 1.88, Windows, and npm checks; fresh checks are required on the final main-based head.
- Both store lifecycle tests passed, covering identity, older writers, allocation, clone/import, restore, runtime cleanup, and malformed or oversized metadata.
- Real CLI fixtures passed for stop/start retention, occupied-address refusal without Gateway/UI startup, and metadata-lock serialization.
- Linux CI passed the affected store, compatibility, output, cleanup, snapshot-restore, and initial UI lifecycle cases. Independent code and test review found no blocker.

The native comparison exercised this patch in the integrated dev workflow against unmodified OpenClaw source. Measured UI ports:

| Native launch | Baseline | Fixed |
| --- | --- | --- |
| Earlier port 5173 occupied | 5174 | 5174 |
| Restart after port 5173 became free | 5173 | 5174 |

The fixed build also refused an occupied retained port before starting Gateway or UI children, preserved the environment identity, configuration and reservation, and recovered at 5174 after release. All completed generations stopped with their listeners closed. Final cleanup removed the test environment and its worktree with no remaining processes.


shakkernerd:
Fixed: dev UI addresses now persist across restarts, with occupied retained addresses refused until available.

PR #207, opened 2026-09-10
feat: request fresh UI links from active dev sessions

shakkernerd:
## What Problem This Solves

Fixes an issue where repeating a matching `ocm dev <env> --ui` command on Linux or macOS reported only the existing address, leaving users without a fresh native owner link after the initial link was consumed or expired.

## Why This Change Was Made

The existing controller now issues a fresh native grant for the requester while retaining ownership of Gateway, Vite and the single dashboard helper. Requests have bounded admission and deadlines; disconnected or late callers cannot send their eventual grant to the original terminal. Stop closes request admission before waiting for verified helper completion.

## User Impact

Matching repeated commands can open the running UI without restarting either component or changing its address. Busy or unready requests report a pending link. Windows retains its existing address-only behavior.

## Evidence

- Focused real CLI cases cover fresh grants with unchanged process identities/address, busy and crashed callers, discarded output, and named stop with pending and full request admission.
- Socket controls cover controller birth identity, filtered replies, framing and reply/ACK bounds, backpressure and abandoned-startup endpoint cleanup.
- Remote Linux: formatting and all-target compilation passed, along with five transport tests, both CLI reuse/lifecycle tests and the dev help aliases. The fresh-reuse case exercises plain and watched sessions.
- [CI passed on `a8e07ac`](https://github.com/openclaw/ocm/actions/runs/34499566197), including the full Linux and native macOS suites, Rust 1.88, Windows compilation, formatting and npm checks. Native macOS passed all five socket tests, both reuse cases and the original UI lifecycle controls.
- The native fresh-grant browser flow passed on the integrated source identified below. Final integration with the effective main dependencies remains a final proof obligation.

## Visual evidence

| Before | After |
| --- | --- |
| Native confirmation/unpaired | Fresh-grant authentication/connected |

### Matching dev invocation and fresh-grant authentication

https://github.com/user-attachments/assets/6560b59c-846d-4da2-9998-f66ca6601ea0

### Visual verification

- Baseline: Same-build native confirmation/unpaired state on a8879f220e6ec184089bc0388eade27a1170e1d6
- Exact head: `a8879f220e6ec184089bc0388eade27a1170e1d6`
- Scenario: In a clean browser, begin at native confirmation/unpaired state, invoke `ocm dev <env> --watch --ui` once against the running session, and authenticate using its fresh native owner link.
[...]

PR #208, opened 2026-09-10
feat: refresh dev UI owner links on Windows

shakkernerd:
Depends on #207.

## What Problem This Solves

Windows reuse only reported the UI address, so a matching `ocm dev` command with `--ui` could not refresh the native browser owner link for an existing session.

## Why This Change Was Made

Connect Windows to the shared handoff path through a private named pipe with current-user access and recorded controller identity checks. One retained pipe instance refuses busy requests and keeps its name reserved between requests. The existing controller retains ownership of each native helper and the Gateway/UI process Jobs.

## User Impact

Matching commands can obtain fresh native grants while keeping the same controller, Gateway/UI processes, and address. Busy or unready requests report a pending link. Disconnected or expired requests cannot send their eventual grant to another terminal.

## Evidence

- Diff hygiene and Node fixture syntax checks pass.
- Remote formatting and locked Windows workspace/all-target compilation passed.
- The shared parent passed eight focused transport, reuse, shutdown, and help checks.
- Windows CI runs the common handoff and named-pipe controls, plus real CLI reuse, busy requests, requester exit, and stop in the existing native Job fixture.
- The native fixture checks process identities, grant ordinals, ownership, and cleanup before its outer Job is closed.


shakkernerd:
Added: Windows dev UI reuse can refresh native browser owner links without restarting the Gateway or UI.

PR #209, opened 2026-09-10
fix: avoid compile-cache wrappers for dev source commands

shakkernerd:
## What Problem This Solves

Fixes an issue where inherited Node compile caching could add an extra launcher process to `ocm dev` and other direct source commands. That wrapper could turn a child's termination signal into an ordinary exit code.

## Why This Change Was Made

OCM now prepares the direct source environment with `NODE_DISABLE_COMPILE_CACHE=1` and removes `NODE_COMPILE_CACHE` before Node starts. Windows environment aliases are handled case-insensitively. This uses the selected OpenClaw checkout unchanged and preserves its native runners, rebuilds, auto-doctor and other `NODE_OPTIONS`.

## User Impact

Direct source commands avoid the compile-cache bootstrap wrapper. Ordinary runtime/launcher settings and the existing managed-service environment policy remain unchanged.

## Evidence

- Remote formatting, locked workspace/all-target compilation and both focused source-environment tests passed. Hosted CI passed on Linux, macOS and Windows; the native Windows lifecycle case checked actual child cache settings and preserved Node options.
- Unmodified OpenClaw `61f7614` launcher files with a controlled application entry were invoked through actual active-source `ocm env exec`. Before, a self-SIGKILL became exit code `1` through an extra cache wrapper. After, the launcher and runtime shared one PID and the parent observed raw `SIGKILL`. Ordinary exit code `23` and `NODE_OPTIONS=--trace-warnings` were preserved.
- Three alternating warm watched-Gateway starts per build, after one warmup each, used the same source, environment, configuration, port and dependency store. Median startup was **6.366s before** and **6.510s after**: **+0.144s (2.3%)** in this small sample. Source and configuration were preserved; cleanup left no owned processes or fixture.


shakkernerd:
Fixed: direct source commands avoid the compile-cache bootstrap wrapper before Node starts.

PR #211, opened 2026-09-10
fix: wait for journal admission in self-update tests

shakkernerd:
## What Problem This Solves

Resolves a problem where the self-update transaction test starts its next operation before the previous worker or activation command has released the update journal lock.

## Why This Change Was Made

Wait for exclusive admission to the existing journal before the same-version no-op and interrupted-update retry. The probe uses the production file-lock mechanism, fails on unexpected I/O errors, and releases its handle before the next CLI invocation. This replaces the orphan activation delay and preserves the busy, unfinished-update, and recovery assertions. Failed no-op and retry assertions now include the command output.

## User Impact

Developers get transaction checks that follow the worker lifecycle. Production update behavior is unchanged.

## Evidence

- Controlled baseline proof distinguishes a terminal update receipt from journal admission: a held lock rejects the next update, and releasing it admits the same no-op.
- `cargo fmt --check` and compilation of the affected integration test passed.
- The existing transaction case passed once on the exact source in 217.67 seconds, including its isolated offline candidate build.
- [Hosted CI](https://github.com/openclaw/ocm/actions/runs/34516044201) passed, including the full Linux/macOS suites, minimum Rust, Windows, and npm checks. The existing transaction case passed on Linux in 199.82s and macOS in 75.24s.
- [CodeQL](https://github.com/openclaw/ocm/actions/runs/34516040062) passed. Independent review is clean, and [ClawSweeper](https://github.com/openclaw/ocm/pull/211#issuecomment-5623752013) rated the exact head Platinum with no findings.

### DIFF
--- a/tests/self_update_transaction_tests.rs
+++ b/tests/self_update_transaction_tests.rs
@@ -9,6 +9,7 @@ use std::thread::sleep;
 use std::time::{Duration, Instant};
 
 use flate2::{Compression, write::GzEncoder};
+use fs2::FileExt;
 use ocm::infra::download::file_sha256;
 use serde_json::{Value, json};
 use support::{TestDir, TestHttpServer, ocm_env, run_ocm_binary, stderr, write_executable_script};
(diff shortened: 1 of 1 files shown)


shakkernerd:
Fixed: self-update transaction tests now wait for journal admission before subsequent no-op and recovery checks.

PR #214, opened 2026-09-11
fix: resolve dev UI paths that use environment variables

shakkernerd:
## What Problem This Solves

Fixes an issue where `ocm dev` left the UI link pending when `gateway.controlUi.basePath` used an environment variable such as `${UI_BASE}`. The Gateway could be healthy at `/console/` while OCM probed the encoded placeholder path.

## Why This Change Was Made

For templated paths, the foreground owner reads only the base path through the selected checkout's native config command, then applies the existing loopback and path validation. OpenClaw owns include and environment resolution, including `.env` and config-provided variables. The resolved target remains captured for the session.

## User Impact

Initial and repeated dev commands can obtain UI links for the effective configured path. Literal paths avoid the extra config command. Failed reads and unresolved or invalid paths produce actionable errors before Gateway and Vite start; query output stays private, and stopping during the read preserves verified cleanup.

## Evidence

- The released-binary baseline requested `/$%7BUI_BASE%7D/` eleven times and received 404 while `/console/` returned 200, leaving both initial and repeated starts without a dashboard handoff.
- Focused CLI coverage exercises includes, caller/config environment precedence, checkout/state `.env` files, selected environment identity, captured-target reuse, private failure output, invalid paths, and cancellation.
- Remote validation passed: `cargo fmt --check`, `cargo check --workspace --all-targets --locked`, the native-config and private failure/cancellation CLI tests, and the existing literal-path fresh-link reuse test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,14 @@ OCM prints the session's UI address, then its initial native owner link after
 both Vite and the Gateway Control UI document are ready. The native handoff keeps
 the Gateway identity and credentials; only the browser document moves to Vite.
 UI requires a local HTTP Gateway with Control UI enabled and installed UI dependencies.
+For a templated `gateway.controlUi.basePath` such as `${UI_BASE}`, OCM reads the
+effective path through the selected checkout's native config command. OpenClaw
+resolves its environment, `.env` files, and config-provided variables; OCM keeps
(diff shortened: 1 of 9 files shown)


shakkernerd:
Fixed: initial and reused dev UI links now use the effective configured base path.

PR #215, opened 2026-09-11
fix: cloned environments load plugins from their source

Marvinthebored:
Related: #98

## What Problem This Solves

Resolves a problem where cloned environments load registered plugins from the source environment instead of their own copy. Cloning must also preserve normally copied local plugins and environments with a missing managed plugin payload.

## Why This Change Was Made

The shared regular and simulation clone path relocates plugin registrations before publishing the clone. It validates the executable installation path against the copied environment and retains original project or archive paths as provenance for managed installations. Both the legacy registry and modern `config_machine_state` representation preserve their other fields and existing wrapper/revision.

## User Impact

Clones use their own plugin files, including linked files owned by the copied environment. Missing managed payloads retain registrations pointing into the clone for diagnosis or repair. Source plugin files and databases stay unchanged, and active paths or writable databases that escape the copied environment are rejected.

This addresses the plugin registry isolation portion of #98.

## Evidence

- All 25 clone tests, 380 library tests, 40 migration tests, and five environment-root isolation tests pass on Crabbox. Coverage includes both registry layouts, directory/archive/npm-pack provenance, missing payloads, owned links and aliases, escaping paths, database links, and simulation clones.
- `cargo fmt --check`, `cargo check --workspace --all-targets --locked`, and `git diff --check` pass.
- Real OpenClaw 2026.9.4 directory and archive installs cloned successfully. `plugins inspect --runtime --json` imported each clone's module and selected its clone-only manifest change. A missing payload remained cloneable, and repair in the target loaded target-owned code. Source database bytes, plugin files, original projects, and archive bytes remained unchanged.
- [Hosted CI](https://github.com/openclaw/ocm/actions/runs/34785373218) also exercises the platform and packaging paths. Its two macOS npm fixture jobs reject an empty `MACOS_TEAM_ID` configuration variable in fork CI before invoking signing tools. The unchanged validator on exact main `1e9b731e3cd70eef565e13771bfcb743f4a707b9` reproduces that argument error. Those jobs test the fixed published v0.2.39 binary; the workflow, fixture, and validator are unchanged by this PR. The separate CI repair is tracked in #216.

All task-owned Rust builds, tests, and live consumer checks ran on Crabbox. No local Rust checks were run.


shakkernerd:
Fixed: environment clones use their own plugin files while preserving copied local installations, missing managed plugin records, and source data.

With thanks to @Marvinthebored.

Issue #217, opened 2026-09-11
Reject overlapping environment roots before creating environment state


shakkernerd:
OCM accepts a new environment root inside another registered environment. Removing the outer environment then deletes the inner environment's state and workspace even when the inner environment is protected, leaving its registration pointing to a missing root. This is data loss through ordinary CLI commands.

New environments must have separate roots: a root must not equal, contain, or sit inside another registered environment's root, regardless of either environment's protection flag. Reject the layout before writing destination state.

Reproduced with OCM 0.2.44 built from [c5ac5392e142f8b8530dd006a627e325e8e4a335](https://github.com/openclaw/ocm/commit/c5ac5392e142f8b8530dd006a627e325e8e4a335) on macOS arm64. `origin/main` still points to that revision as checked on 2026-09-11. Put that executable on `PATH`; this reproduction uses temporary isolated state and starts no services:

```sh
ocm_bin="$(command -v ocm)" || exit 1
fixture="$(mktemp -d)" || exit 1
env -i PATH="$PATH" OCM_BIN="$ocm_bin" FIXTURE="$fixture" sh <<'SH'
set -eu
trap 'rm -rf "$FIXTURE"' 0
export HOME="$FIXTURE/home"
export OCM_HOME="$FIXTURE/ocm-home"
export TMPDIR="$FIXTURE/tmp"
mkdir -p "$HOME" "$OCM_HOME" "$TMPDIR"

"$OCM_BIN" env create outer
inner="$OCM_HOME/envs/outer/nested"
"$OCM_BIN" env create inner --root "$inner" --protect
printf 'protected child data\n' > "$inner/.openclaw/workspace/keep.txt"

"$OCM_BIN" env create sibling --protect
sibling_marker="$OCM_HOME/envs/sibling/.openclaw/workspace/keep.txt"
printf 'protected sibling data\n' > "$sibling_marker"

if "$OCM_BIN" env remove inner; then
    exit 1
fi
test -f "$inner/.openclaw/workspace/keep.txt"

"$OCM_BIN" env remove outer --json
test ! -e "$inner"
"$OCM_BIN" env show inner --json
"$OCM_BIN" env doctor inner --json
cat "$sibling_marker"
SH
```

Observed results:

- Direct `env remove inner` exits 1 with a protection error and preserves the child marker.
- `env remove outer --json` exits 0 without `--force`. The child's root and marker disappear.
[...]

PR #218, opened 2026-09-11
fix: reject overlapping environment roots

shakkernerd:
Related: #217

## What Problem This Solves

Fixes an issue where users creating an environment inside another environment's root could lose its state and workspace when removing the parent, even when the child was protected.

## Why This Change Was Made

Create, clone, and import now reject equal, ancestor, or descendant root overlap while holding the existing environment registry lock, before creating or copying destination state. The shared check resolves existing and missing paths and includes symlinks needed to reach registered roots. Start/setup, dev, migration/adopt, and simulation use these guarded admission paths.

## User Impact

The error identifies the conflicting environment and root. Ordinary sibling roots and disjoint custom roots remain valid, regardless of protection flags. Existing dev-source safeguards remain intact. This changes new admissions only; it adds no repair or deletion handling for pre-existing nested layouts.

README, usage guidance, and the affected command help describe the rule.

## Evidence

- The retained macOS arm64 reproduction at `c5ac5392` demonstrates protected child data loss through ordinary create/remove commands.
- Formatting, diff checks, all-target compilation, and 160 affected tests passed on Linux with Rust 1.88.0. Coverage includes actual create/clone/import commands, simulation, setup/start/dev/migrate/adopt callers, help, aliases, missing roots, disjoint controls, and dev-worktree recovery.
- An isolated direct CLI check rejected the protected nested root before creating it or changing registry bytes. Removing the outer environment then preserved the protected sibling and its marker.
- Independent source review, including the recovery-fixture correction, found no blockers.
- Hosted Linux/macOS tests, Windows compilation, minimum-Rust validation, and the current automated review are tracked in this PR's checks. Windows runtime behavior has not been executed.


shakkernerd:
Fixed: New environment roots are rejected before destination writes when they overlap a registered environment, including aliases and missing roots.

PR #225, opened 2026-09-13
fix: keep resolved release details in upgrade reports

### LINKED ISSUE #127: Upgrade failure summaries discard resolved release-channel metadata
## Summary

When an exact-version upgrade fails, OCM can report:

```json
{
  "runtimeReleaseVersion": "2026.8.1",
  "runtimeReleaseChannel": null,
  "outcome": "failed"
}
```

OCM had already resolved that release and knew its channel was `stable`. The preceding dry run correctly reported:

```json
{
  "runtimeReleaseVersion": "2026.8.1",
  "runtimeReleaseChannel": "stable",
  "outcome": "would-switch"
}
```

## Reproduction

Run an exact-version upgrade whose candidate preparation fails:

```bash
ocm upgrade test-env \
  --version 2026.8.1 \
  --json
```

The resulting failure summary retains the resolved version but sets `runtimeReleaseChannel` to `null`.

## Actual behavior

Successful and dry-run summaries use the resolved target’s release metadata.

Failure and rollback paths instead derive the channel from the original CLI arguments. Therefore:

- `--channel stable` retains `stable`.
- `--version 2026.8.1` loses the resolved `stable` channel.
- `--runtime <name>` can lose channel metadata recorded on that runtime.

## Expected behavior

Once OCM resolves an upgrade target, all summaries for that operation should retain the same resolved version and channel, regardless of whether preparation, commit, post-update work, or rollback succeeds.

## Root cause

`ResolvedUpgradeTarget` contains the resolved `release_channel`, and the dry-run path uses it directly.

The preparation path consumes that resolved target. Error paths then call `release_channel_hint()`, which only returns an explicitly supplied `--channel` value. It cannot recover release metadata resolved from `--version` or `--runtime`.

## Impact

JSON consumers receive inconsistent target identity depending on the upgrade outcome. This weakens diagnostics, audit records, automation, and retry decisions because a failed operation no longer identifies the release channel OCM actually selected.

## Required fix

Preserve the resolved target metadata for the complete upgrade operation and use it for every success, failure, and rollback summary.

Do not reconstruct resolved metadata from the original CLI selector after resolution.

## Regression coverage

Add coverage proving that:

- Exact-version preparation failures retain the resolved channel.
- Exact-version rollback summaries retain the resolved channel.
- Named-runtime failures retain channel metadata from the runtime.
- Runtime-bound and launcher-bound environments behave consistently.
- Explicit `--channel` behavior remains unchanged.
- Targets with genuinely unknown channel metadata continue reporting `null`.

shakkernerd:
Closes #127

## What Problem This Solves

Fixes an issue where users upgrading by version or runtime name could receive failure or rollback reports that omitted a known release channel. Named runtimes could also lose their detected version in the final report, and failed manifest upgrades could report the previous version instead of the selected target.

## Why This Change Was Made

Keep the resolved target version and channel throughout the upgrade and use them consistently when reporting its result. Preserve unknown channels as `null` and known custom version labels such as `nightly`, while leaving release selection, version comparison, installation, and recovery behavior unchanged.

## User Impact

Upgrade reports identify the same selected release across previews, success, preparation failures, and automatic rollback, including launcher-based environments and the `latest` alias for `stable`.

## Evidence

- Unchanged main reproduced missing channels in exact-version preparation and early rollback reports, a missing detected version after named-runtime success, and the previous version after failed manifest preparation.
- Exact-head Crabbox validation passed `cargo fmt --check`, `cargo check --workspace --all-targets --locked`, 15 upgrade unit tests, and 119 affected integration tests: **134 tests passed**.
- Real CLI fixtures cover version/runtime selectors, launcher recovery, unknown channels, custom labels, alias normalization, and running/stopped service restoration.
- [Platform CI passed](https://github.com/openclaw/ocm/actions/runs/34760237867), including full Linux/macOS suites, installation smoke tests, minimum Rust, Windows, and native macOS cleanup controls.


shakkernerd:
Fixed: upgrade reports preserve resolved release details across previews, success, failures, and automatic rollback.

PR #231, opened 2026-09-15
fix: prevent empty PID reads in dev UI tests

shakkernerd:
## What Problem This Solves

Resolves intermittent dev UI test failures where the dashboard PID file is empty when a waiting test reads it. This caused [main CI to fail while checking session reuse](https://github.com/openclaw/ocm/actions/runs/34832249350/job/103938022491).

## Why This Change Was Made

The fixture now finishes writing each PID to a sibling temporary file before renaming it to the path readers wait for. A regression test observes the actual open-before-write interval and verifies that the final path appears only with the complete child PID.

## User Impact

Dev UI tests get reliable PID readiness. OCM commands and their behavior are unchanged.

## Evidence

- The actual baseline fixture failed the deterministic regression with an empty final PID file; the repaired fixture passed and published the correct PID.
- Crabbox: `cargo fmt --check` and `cargo test --locked --test dev_command_tests dev_ui_dashboard_publishes_complete_pid -- --exact --nocapture` passed.
- Crabbox: `cargo check --workspace --all-targets --locked` passed; `cargo test --locked --test dev_command_tests` passed all 84 tests, including the previously failing session-reuse test.
- Hosted platform coverage: [CI run](https://github.com/openclaw/ocm/actions/runs/34911804013).


shakkernerd:
Fixed: dev UI tests no longer read partially written dashboard PID files.

PR #234, opened 2026-09-15
chore: group routine GitHub Actions updates

shakkernerd:
## What Problem This Solves

Routine GitHub Actions updates currently arrive as a separate Dependabot pull request for each dependency. When several minor or patch releases become available together, maintainers have to repeat the review and CI cycle across multiple small PRs, making the update queue harder to manage.

## User Impact

Eligible minor and patch Actions version updates will share one Dependabot group. Major Actions upgrades remain separate, and Cargo updates continue through their existing configuration so runtime dependency changes retain their own review.

The weekly schedules, five-open-PR limits, and existing security-update behavior are preserved. Grouping reduces routine PR volume; it does not eliminate update PRs or automatically approve or merge them.

## Why This Change Was Made

The existing `github-actions` entry now uses Dependabot's native `groups` option with `applies-to: version-updates`, a wildcard dependency pattern, and only `minor` and `patch` update types. The wildcard includes Actions from all publishers, while the update-type restriction keeps major upgrades outside the routine group.

This uses the existing update service without adding filters, workflows, dependency changes, or release behavior. The rule follows GitHub's [documented grouping options](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups).

## Evidence

The system YAML parser successfully loaded the updated file. A structural comparison against the base revision confirmed that removing the new group produces the original configuration exactly, covering the Cargo entry, schedules, limits, and all other settings. The focused preflight also verified the group fields and passed `git diff --check`; it completed in 0.110 seconds.

The complete contribution changes only `.github/dependabot.yml`. All nine jobs in the [CI run](https://github.com/openclaw/ocm/actions/runs/35036353495) and all CodeQL checks passed on the final commit. Independent review found no issues. This is static configuration proof; a future Dependabot update run will demonstrate the resulting grouped PR in operation.


shakkernerd:
Improved: routine minor and patch GitHub Actions version updates now share one Dependabot group.

PR #131, opened 2026-08-31
fix: isolate runtime install lifecycle state

MertBasar0:
Related: #98

## What Problem This Solves

Resolves a problem where preparing an OpenClaw runtime lets npm installation scripts discover and modify the caller's existing OpenClaw state before the new runtime is recorded.

## Why This Change Was Made

Package and companion installation now direct OpenClaw's home, state, and config selectors into disposable installation state and clear inherited service-state and active-environment selectors. The npm process retains the caller's home, configuration, and script policy. Installation state is removed when npm returns, including a normal failure.

## User Impact

Preparing a runtime preserves caller OpenClaw state through the supported lifecycle path selectors. User npm settings, including disabled scripts, explicit configuration files, and paths expanded from HOME, continue to apply.

## Evidence

Validation on Crabbox with Rust 1.88.0, Node.js 24.18.1, and npm 11.16.0:

- Formatting, locked all-target compilation, JavaScript/shell syntax, and complete Rust test coverage passed, including 56 runtime tests and 40 service tests.
- Real npm honored enabled/disabled scripts with both default and explicit configuration files, including an explicit file outside HOME. Enabled hooks retained HOME, USERPROFILE, effective userconfig, and HOME-expanded cache settings.
- An intentionally failing hook produced no published runtime or installation residue; caller database, config, and npmrc bytes stayed unchanged.
- The unchanged beta postinstall cleanup code executed with and without an inherited service-state selector. Both caller dependency sentinels remained intact.

The reproduction uses the unmodified schema-6 database created by published OpenClaw `2026.8.1-beta.1` and installs `2026.8.1-beta.2`. Database inspection is read-only, and the explicit-path control uses the legacy installer.

| Installer | Database schema | Database bytes | Service-state sentinel |
| --- | --- | --- | --- |
| OCM v0.2.32 | 6 → 8 | Changed | Removed |
| Fixed OCM | 6 → 6 | Preserved | Preserved |
| OCM v0.2.32 with explicit disposable paths | 6 → 6 | Preserved | Preserved |

Config bytes remained unchanged in all three cases. The [hosted reproduction](https://github.com/MertBasar0/ocm/actions/runs/34829551478) also passed.

All [required CI jobs](https://github.com/openclaw/ocm/actions/runs/34829551463) passed. The two optional macOS npm jobs reproduce the unchanged-main empty team-ID failure tracked by [#216](https://github.com/openclaw/ocm/pull/216).

PR #235, opened 2026-09-15
chore: remove the standalone OCM-98 workflow

shakkernerd:
Related: #131.

## What Problem This Solves

OCM still carries a dedicated GitHub Actions workflow for the historical runtime-install isolation investigation in #98. It runs manually or on one reproduction branch, repeats formatting and test checks covered by normal CI, builds current and older OCM binaries, and uploads the comparison results.

Keeping that experiment as a separate workflow also gives Dependabot another set of action versions to maintain. The checkout, Rust-cache, and artifact-upload updates in #230, #229, and #227 are confined to this file. The runtime isolation regression already belongs to the normal test suite, while the standalone reproduction script remains available when a historical comparison is useful.

## User Impact

This removes the standalone OCM-98 workflow definition from the default branch. Maintainers retain the runtime-isolation tests, the reproduction script and its database helper, and the existing diagnostic evidence. A deliberate reproduction can still use the script with the selected current and legacy OCM binaries.

Normal pull-request CI, release validation, signing and notarization, and npm publication continue through their existing workflows. OCM commands, installed environments, dependency versions, and release behavior are unaffected by this maintenance change.

## Why This Change Was Made

The ongoing requirement is to preserve the runtime-isolation repair and its regression coverage. Normal CI already runs the locked Rust test suite containing that regression; maintaining a second GitHub workflow is not needed to keep the repair covered.

The change therefore removes only the standalone workflow declaration. The source tree has no external references that require a replacement or documentation rewrite. The diagnostic script and database helper remain unchanged, and no new workflow, runner, package script, or dependency configuration takes the deleted file's place.

## Evidence

Static validation confirms that the complete change is one file deletion. All 214 remaining tracked entries retain their exact Git objects and file modes relative to the current base, including the runtime implementation and tests, both reproduction helpers, every remaining workflow, and the Cargo and Dependabot configuration. A tracked-source reference scan found no dangling workflow filename, workflow name, or dedicated-branch references.

[CI on the final commit, `3811190`](https://github.com/openclaw/ocm/actions/runs/35040221124) passed all nine jobs, including the full Linux/macOS suites, native installation smoke checks, Windows and minimum-Rust checks, formatting, and npm verification. The retained runtime-install isolation regression explicitly passed in the [Linux job](https://github.com/openclaw/ocm/actions/runs/35040221124/job/104618196435) and [macOS job](https://github.com/openclaw/ocm/actions/runs/35040221124/job/104618196656). All four [CodeQL analyses](https://github.com/openclaw/ocm/actions/runs/35040217893) also passed.

Independent review found no issues in the complete contribution. Historical diagnostic proof from #131 remains available as historical evidence; removing this workflow did not require rebuilding the old binary or repeating that comparison. No local Rust build, test, or native execution was performed for the static preflight, and no additional reproduction runner was created.


shakkernerd:
Simplified: removed the standalone OCM-98 workflow while preserving normal CI coverage and the diagnostic scripts.

PR #239, opened 2026-09-16
chore: remove obsolete OCM-98 reproduction tools

shakkernerd:
## What Problem This Solves

The standalone OCM-98 reproduction workflow was removed in #235, but its incident-specific shell script and SQLite metadata helper remained in the repository.
The reproducer requires a legacy OCM v0.2.32 binary and fixed OpenClaw beta fixtures, and neither tool is used by the current runtime, CI, release, packaging, or documented development workflow.
Keeping these obsolete entry points alongside maintained tooling leaves an unnecessary historical workflow in the source tree.

## User Impact

This removes `scripts/reproduce-ocm-98.sh` and `scripts/ocm-98-db-meta.mjs` together.
The historical investigation remains available in Git history.
OCM's runtime-install behavior, caller-state isolation, npm configuration handling, and normal test workflow remain unchanged.

The runtime-isolation fix from #131 stays in `src/store/runtimes.rs`, together with the active regression tests and their support probes.
README and CONTRIBUTING guidance, normal CI, and all release, signing, installer, and npm tooling are preserved.

## Why This Change Was Made

The database helper is used only by the obsolete reproducer, so deleting the pair removes the complete unused tooling without leaving another orphaned helper.
Current source and discovery paths have no other caller for either file.
The actual protection against the original regression belongs to the runtime implementation and its ordinary tests, which do not depend on these historical tools.

## Evidence

The static preflight verified an exact two-file deletion: every remaining tracked object and mode matches the current base, and no retained source reference points to either removed file.
Independent review also checked indirect discovery, explicit release fixture lists, and native/npm package contents and found no dependency on the pair.

The preserved `official_runtime_install_isolates_npm_lifecycle_from_caller_openclaw_state` test covers caller state and service-path isolation through host and managed npm paths.
The preserved `official_runtime_install_preserves_real_npm_configuration` test covers script policy and normal or explicit npm configuration.
Both remain in the existing Linux/macOS `cargo test --locked` CI path.
The obsolete reproducer and legacy binaries were not executed, and no local Rust build or dependency installation was needed for this deletion-only check.
[CI run 35086808704](https://github.com/openclaw/ocm/actions/runs/35086808704) passed all nine jobs for commit `4a4a508`, including both complete test suites and `cargo install` smoke checks.
All four [CodeQL analysis jobs](https://github.com/openclaw/ocm/actions/runs/35086805374) passed.
The [completed automated review](https://github.com/openclaw/ocm/pull/239#issuecomment-5696265797) records Platinum (4/6) with no findings for the same commit.


shakkernerd:
Simplified: OCM's source tree no longer includes the obsolete OCM-98 reproduction tools.

Issue #240, opened 2026-09-16
Adoption rewrites checkpointed audit archives and blocks Doctor recovery


shakkernerd:
## What happens

Adopting an existing OpenClaw home with `ocm adopt import` can change the contents of its already-migrated audit archives. The adoption command succeeds, but a later OpenClaw Doctor or update finalization can refuse audit recovery because the copied archive no longer matches its checkpointed content hash.

This affects historical records containing the old state-directory path, such as the `configPath` recorded in a configuration audit entry. The original source files remain intact; the imported copy is changed.

## Reproduction

The following creates only temporary synthetic state and uses the normal adoption command. It does not start a Gateway or modify a database or checkpoint.

```python
import hashlib
import json
import os
from pathlib import Path
import subprocess
import tempfile

with tempfile.TemporaryDirectory(prefix="ocm-audit-adopt-") as temporary:
    root = Path(temporary)
    source = root / "old" / ".openclaw"
    (source / "logs").mkdir(parents=True)
    (root / "home").mkdir()
    (source / "openclaw.json").write_text("{}\n")
    relative = Path("logs/config-audit.jsonl.migrated")
    original = (json.dumps({
        "event": "config.write",
        "configPath": str(source / "openclaw.json"),
    }) + "\n").encode()
    (source / relative).write_bytes(original)
    env = {
        "PATH": os.environ["PATH"],
        "HOME": str(root / "home"),
        "OCM_HOME": str(root / "ocm-home"),
    }
    imported = json.loads(subprocess.check_output([
        "ocm", "adopt", "import", "--name", "audit-repro",
        str(source), "--json",
    ], env=env, cwd=root, text=True))
    copied = (Path(imported["root"]) / ".openclaw" / relative).read_bytes()
    print("archive preserved:", copied == original)
    print("checkpointed prefix preserved:",
          hashlib.sha256(copied[:len(original)]).digest()
          == hashlib.sha256(original).digest())
    subprocess.run([
        "ocm", "env", "destroy", "audit-repro", "--yes",
    ], env=env, cwd=root, check=True, stdout=subprocess.DEVNULL)
```

Expected: both values are `True`. With OCM 0.2.43 and a native build of current main (0.2.47 at `a2784525`), both are `False`, although `adopt import` returns exit code 0.

[...]

PR #241, opened 2026-09-16
fix: preserve audit archives when adopting OpenClaw homes

shakkernerd:
Related: #240

## What Problem This Solves

`ocm adopt import` and `ocm migrate` can rewrite historical paths inside an existing OpenClaw audit archive while copying it into a managed environment. Adoption succeeds, but the changed archive no longer matches its existing content checkpoint, so a later Doctor or update finalization can refuse audit recovery. The same historical paths can also be misreported by OCM as copied runtime references requiring repair.

## User Impact

Adoption preserves the original bytes of OpenClaw's legacy configuration, system-agent, and Crestodian audit families, including their migration and recovery artifacts. Historical paths remain valid evidence for OpenClaw's checkpoint checks, and OCM no longer classifies them as runtime-path drift.

Normal environment-scoped references in configuration, session state, application logs, and config backups continue to relocate. OpenClaw retains ownership of audit recovery and checkpoint updates; environment clone/export/import retention is unchanged. Existing imports affected by an older version still need recovery from their unmodified source.

## Why This Change Was Made

The generic relocation walker treated every readable text file as mutable runtime state. The correction adds one state-root-relative classifier for the audit families and uses it in both relocation and copied-reference inspection. It recognizes supported numbered generations without imposing an integer-size limit, while unrelated filenames and matching filenames outside the owning directories keep their existing behavior.

The README explains why these historical paths remain unchanged. The correction requires no database, checkpoint, dependency, configuration, or service-lifecycle changes.

## Evidence

The native OCM 0.2.43 release and an unmodified build of main at `a2784525fc4cb1e37b68ec9905a13c46a6f2cd19` both reproduced the issue with tiny synthetic Linux fixtures. All three audit-family samples changed exactly by the state-root substitution and failed their original prefix digest, while the original files remained intact. See #240 for the standalone reproduction.

The added adoption regression fails on the original implementation and passes with the correction. It checks byte preservation for 33 supported audit paths, five unrelated or malformed lookalikes, and the existing ordinary-relocation controls. The extended state-audit test verifies historical audit paths are ignored while an ordinary foreign runtime reference still produces the existing finding.

Completed on the retained Linux runner:

- `cargo fmt --check`
- `cargo test --locked --lib store::openclaw_state::tests`
- `cargo test --locked --test migrate_command_tests --test env_doctor_tests --test env_cleanup_tests --test env_snapshot_tests`

All 97 affected tests passed. A fresh native adoption with the fixed binary also preserved all three synthetic archives and their prefix digests while the ordinary application-log reference relocated. Changed file hashes matched the published head.

Independent engineering review found no blocking defect. Hosted CI passed on `b68220bf9f10514fbef0812135e591fc0e1511d0`, including Linux and macOS tests, Rust 1.88 compatibility, Windows compilation, and npm checks. CodeQL also passed. The command proof establishes Unix adoption behavior; Windows evidence is compilation-only. Byte preservation leaves OpenClaw's separate recovery-journal identity checks intact and does not promise every interrupted recovery is resumable after relocation.


shakkernerd:
Fixed: OpenClaw adoption preserves legacy audit archives and avoids misleading runtime-path repair warnings.

